PR #8, opened 2020-06-11
New Hue firmware files

mkaay:


Koenkk:
Thanks!

Tone866:
the json is not valid anymore now @Koenkk .
In your index.json the "path" parameter is missing in the later objects, so line 62 is not valid anymore because the "," at the end.
mkaays json instead is valid.

So updates now run in an error: 
zigbee2mqtt:error 2020-06-12 02:30:07: Failed to call 'OTAUpdate' 'onZigbeeEvent' (TypeError: images.find is not a function

Koenkk:
@Tone866 fixed!

PR #14, opened 2020-10-29
Securifi Peanut firmware

walthowd:
This SecuriFi peant plug seems to be popular in the US, and this firmware file adds power monitoring to the plug.

Getting the plug to request the firmware is a bit odd, and may require some changes in zigbee-herdsman. 

To initiate the OTA transfer you have to write attribute 2 (groups) to cluster 5 (scenes) with the value 49502

Thanks to @tube0013 for sniffing the traffic from the OEM hub and figuring this out here:

https://github.com/zigpy/zigpy/issues/522#issuecomment-718815257


Koenkk:
Thanks!

PR #20, opened 2021-01-27
Added new Philips Hue firmware images

TheJulianJES:
This adds new Philips Hue images (mostly for the Bluetooth lamps/plugs).

@mkaay If you have some spare time, would you mind checking for any other new OTAs using your script?


mkaay:
Hey Julian, yes I managed to find a few new versions.
Will add them after your pull request gets approved so we won't get conflicts.
I updated my script to also check the Bluetooth API. Looks like the Bridge image versions are almost the same, only 4 updates are missing this time.

Koenkk:
Thanks!

PR #23, opened 2021-02-03
Adds OTA image for GL-S-007P

pedrolamas:
This update is a follow up on https://github.com/Koenkk/zigbee2mqtt/discussions/2921#discussioncomment-331372

I've also taken the opportunity to update the `scripts/add.js` to ensure one can pass a `modelId` as the 3rd parameter (necessary for these Gledopto devices)


Koenkk:
Thanks!

PR #24, opened 2021-02-03
Fixes url escaping

pedrolamas:
Using the add.js will set a unescaped url that won't actually work to retrieve the file, this fixes that.


Koenkk:
Thanks again!

PR #25, opened 2021-02-04
Add sha512 checksum to OTA firmwares

davidjb:
First part of https://github.com/Koenkk/zigbee-herdsman-converters/issues/1998

This hashes firmware during download and adds the hash to index.json; sha512 follows NPM's lead for package validation.

Additionally, this PR adds an `updateall.js` script which reads all entries in `index.json` and runs `add.js` on each of them (for mass updates like the addition of an attribute or change).  To allow for this to happen correctly, this PR adds the ability to modify entries in-place via `add.js` (if called on an existing path in the repo -- previously it tried to unlink/copy the firmware onto itself).

Lastly, this normalises the paths for the Gledopto firmware to lowercase for its file extensions; they were previously `.OTA` in index.json but `.ota` in the repository & in the `url` field -- this was something that got picked up by the mass update.


Koenkk:
Thanks!

PR #28, opened 2021-03-10
Xiaomi/Aqara OTA-firmware

kirovilya:


Koenkk:
Thanks, for this to work you also need to add `ota: ota.zigbeeOTA,` (e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/729afa44554aa54e1a30d27207208757cbbb50dc/devices.js#L1630). Could you also make a pr for that?

kirovilya:
hmm. ok. later

svrooij:
I’m curious where you got the files? Does this also mean I will soon be able to update my Xiaomi buttons as well

kirovilya:
@svrooij not for all devices
@Koenkk did it

svrooij:
I have a lot of [these](https://www.zigbee2mqtt.io/devices/MCCGQ01LM.html#xiaomi-mccgq01lm) and [these](https://www.zigbee2mqtt.io/devices/WXKG01LM.html#xiaomi-wxkg01lm) and I also have a xiaomi gateway collecting dust somewhere.

Can I set it up to sniff the location of the firmware files?

VOVAiS:
Don’t work lumi.ctrl_86plug.aq1 after update. Blink blue and don’t work

mSapik:
After update lumi.light.aqcn02 from 1.23 to new firmware, lamp don't work. I try delete lamp and join, after switch the lamp on five times, lamp blinking without stopping. [https://youtu.be/QCwVXEpwQFw](url)

kirovilya:
@VOVAiS hmm. I do not know yet. I unfortunately did not check these firmware - they got to me as is.
I ordered myself an aqara bulb to check the ota.

VOVAiS:
@kirovilya, @Koenkk 
unfortunately I think we need to remove OTG support for Xiaomi. There will be many broken devices

kirovilya:
make PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/2374 
until checked

VOVAiS:
@kirovilya  Is the question now is it possible to recover?
there is a chip jn5163
I have other outlets. Is there a way to read the firmware and write?
Is there a way to read the firmware and write?

kirovilya:
@VOVAiS I don't know

nukusinji:
My curtain motor do not work any more after I updated ZNCLDJ11LM (OTA_lumi.curtain_0.0.0_0034_20200717_1F0A7B.20200717154021.ota).
I try to repairing, remove is ok. but device do not enter pairing mode.
other motor is only device announce when replug power.

I wanna restore firmware. How about it?

VOVAiS:
Connected  on the UART.
The problem is that the firmware is protected from reading. Will not download firmware from a good device

VOVAiS:
You can restore via uart, but you need firmware files in .bin.
uploaded the firmware https://github.com/igo-r/Lumi-Router-JN5169. the outlet has become GWRJN5169 device. everything upset me

nukusinji:
> You can restore via uart, but you need firmware files in .bin.
> uploaded the firmware https://github.com/igo-r/Lumi-Router-JN5169. the outlet has become GWRJN5169 device. everything upset me

Did you find original firmware file?
I can not find it. so I ordered new two curtain motor.

VOVAiS:
No, I could not find it. I found a firmware from another device  (GWRJN5169). The socket became just a signal repeater. Without the original firmware of the device in the trash

kirovilya:
@VOVAiS can you write to me in telegram? https://t.me/goofyk

kirovilya:
@nukusinji can you write to me in telegram? https://t.me/goofyk

VOVAiS:
Thank you @kirovilya. I was given the firmware in the bin file and I restored my device

nukusinji:
> Thank you @kirovilya. I was given the firmware in the bin file and I restored my device

Thanks for you information.

Can I have question.
How to write the bin file?
 This link(https://github.com/igo-r/Lumi-Router-JN5169) method is connect SSH.. 

I just have the broken device and 'serial to USB' converter. So I can connect PC and the device through USB.
If I can write with my tool, Which SW can I use?

VOVAiS:
@nukusinji 
you can use usb uart
You need to solder the wires like here https://github.com/Jerome-PS/JN516xSniffer. You need
Purple: Tx 
Green : Rx 
Bleue : nBootloader on GND
Red : 3V3
Black : GND 
Soft JN51xx Production Flash Programmer software

PR #38, opened 2021-04-04
Xiaomi/Aqara OTA (take two)

kirovilya:
https://github.com/Koenkk/zigbee2mqtt/issues/6906


Koenkk:
Thanks!

sjorge:
Still seems to be something wrong, after updating to latest z2m dev ota can now be checked for https://www.zigbee2mqtt.io/devices/GZCGQ01LM.html#xiaomi-gzcgq01lm (only device I have paired at the moment), I’d except it to say there is now update available. But instead it hits and undefined variable error.

`Failed to check if update available for 'balcony/lux' (Cannot read property 'ID' of undefined)`

I’ll try to dig into this a bit next time I am not on my tablet.

sjorge:
Still seems to be something wrong, after updating to latest z2m dev ota can now be checked for https://www.zigbee2mqtt.io/devices/GZCGQ01LM.html#xiaomi-gzcgq01lm (only device I have paired at the moment), I’d except it to say there is now update available. But instead it hits and undefined variable error.

kirovilya:
@sjorge  It turns out that this sensor does not have an OTA cluster. We need to learn how to update the firmware on them ... while I turn off the update for it.

sjorge:
> @sjorge It turns out that this sensor does not have an OTA cluster. We need to learn how to update the firmware on them ... while I turn off the update for it.

Those are pretty horrible firmware wise too:
image

Yes, that is really what the version returns, seems like some wrong buffer. I have it on all 3 of mine and they were ordered at seperate times months apart.

PR #47, opened 2021-08-12
Update Hue firmware images (breaking?)

TheJulianJES:
These firmware images do not seem to be rolled out fully. ~~They aren't provided through the Hue BT app for example.~~
These updates are now (18/08/2021) also rolling out through the Hue Bluetooth app.

Version code for the older ZLL (non-bluetooth) lights seems to be: 1.88.1 / 20210331
(Previous firmware for most Hue ZLL lights: 1.50.2_r30933 / 20191218)

Version code for the newer ZHA (Bluetooth) lights seems to mostly be: 1.90.1 / 20210525
(Previous firmware for most Hue ZHA lights was 1.7.10 or 1.7.11)

The manufacturer on the older ZLL lights was now also changed from "Philips" to "Signify Netherlands B.V."
(It's likely that the same is done on the rest of the newer" Zigbee 3.0 lights with Bluetooth".)
Since this might require changes in Z2M and ZHA (and I'm not sure if these images are even rolled out yet), it's a draft PR for now.
Changes "required" for ZHA:
- https://github.com/home-assistant/core/pull/54513
- https://github.com/zigpy/zigpy/pull/788
- https://github.com/zigpy/zha-device-handlers/pull/993

Signatures (ZHA):
- ZLL profile after re-pair: https://paste.ubuntu.com/p/RKbpqqD4Rt/
- ZHA profile before re-pair: https://paste.ubuntu.com/p/hM2HGFVqkT/
- ZHA profile after re-pair: https://paste.ubuntu.com/p/D87NfZcFQ8/

For the older (ZLL profile) lights, it looks like the manufacturer name changed (and the 0xfc03 cluster was added).
For the newer (ZHA profile/Bluetooth) lights, it looks like cluster 64514/0xfc02 changed to 64515/0xfc03 (looks like some lights now have both clusters?).
I don't think the cluster change will cause any issues with Z2M?

Changes: https://hueblog.com/2021/08/14/firmware-update-prepares-hue-lamps-for-dynamic-scenes/

I'll probably wait until marking this "as ready" until the next Home Assistant release is out (as ZHA only then fully supports Hue's default startup behavior without needing any quirks and includes the increased polling rate for older, updated lights).

I'll update this PR as I get more information about this.

### DIFF
--- a/index.json
+++ b/index.json
@@ -63,28 +63,28 @@
         "path": "images/Sengled/RDS2017028_E1C-NB6_V0.0.22_20180314.ota"
     },
     {
-        "fileVersion": 1107322837,
+        "fileVersion": 1124096001,
         "fileSize": 258104,
         "manufacturerCode": 4107,
         "imageType": 256,
-        "sha512": "7e1618bbad65398d4a7d07d57bef3e443d21869a9ff86f24316fe0b018593c4c03001ee0abce2fe9afe182b83ca72cf551733f7cc38e8baf1fef9107e182485e",
-        "url": "http://fds.dc1.philips.com/firmware/ZGB_100B_0100/1107322837/TI_0100_5.127.1.26581_0012.sbl-ota"
+        "sha512": "3ed48f3f754726db044efb03b74bda21eb0f85ad7814d43d3011ef84b4b1bc48f87b02d01bc926553eb746063e3e34d1e2ab7c9423ece92312af73b3313a7401",
+        "url": "http://fds.dc1.philips.com/firmware/ZGB_100B_0100/1124096001/Ti_0100_ConnectedLamp-TI-Target_0012_88.1.sbl-ota"
     },
     {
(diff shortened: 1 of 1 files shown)


sjorge:
I upgraded my GU10 via the BT app and it still worked, so it probably only matches on model so it might be fine.

sjorge:
Also dug up my old E14 non BT one, update went fine and it still worked. Was able to remove and join it without issue:
image

TheJulianJES:
The latest Home Assistant version is now fully compatible with these updates. Should be ready for merging.

Koenkk:
Thanks!

jeppper:
I am on 1.18.1 because I use HA 0.103.6 and am having big problems with strange behaviour of my Philps Hue lights. I updated some weeks ago via OTA recently to 1.88.1 and 67.88.1.
Could there be a problem using the new Hue Firmware with Z2M 1.18.1?

PR #49, opened 2021-09-01
Adds Xiaomi Plug OTA v30

AlexxIT:
For model `lumi.plug`. Because you can't update plug to latest fw if they have a firmware version below 30.
You need update to v30 fw and only then to v90:
https://t.me/xiaomi_gw_v3_hack/49948


Koenkk:
This image won't be used if it is not in https://github.com/Koenkk/zigbee-OTA/blob/master/index.json

AlexxIT:
I know this. But someone can add two step update logic after this pull request.

Koenkk:
That should be added first then, otherwise it makes no sense to merge this pr.

u17194907425:
@Koenkk good day. I need your advice. If I go ahead and do the following

1. update the https://github.com/Koenkk/zigbee-OTA/blob/master/index.json file using the procedure described here https://github.com/Koenkk/zigbee-OTA to point at the v30 FW 
2. Temporarily replace this file on my local HASS installation 
3. update all the plugs using standard zigbee2mqtt mechanisms
4. roll-back `index.json` changes

does it have a potential to work or do I miss something basic preventing this idea from being successful? In case the concept looks right can you suggest what is the right way do the step number 2?

Thanks

PR #59, opened 2021-11-11
Fix image parsing when the upgrade file identifier is not at position 0

cemonneau:
Hi,

On some OTA firmwares, the upgrade file identifier ([0x1E, 0xF1, 0xEE, 0x0B]) doesn't start at 0. z2m has the right code to detect it but not this repository which prevent to detect some firmwares properly : in my case some Legrand that I'm working on to test for a future integration.


Koenkk:
Thanks!

PR #71, opened 2021-12-14
Update QBCZ14LM OTA file

z-x-f:
Version: 50


Koenkk:
To be on the safe side, please also add the `modelId` like https://github.com/Koenkk/zigbee-OTA/blob/7f3800dfd5995bbb9cb1ec62ecccf879e3b6a9c2/index.json#L879

PR #78, opened 2022-01-02
Firmware for EasyAccess/Home Control/Wattle Door Locks

heinekmadsen:


index.json
```diff
@@ -1408,5 +1408,14 @@
         "modelId": "lumi.plug.sacn03",
         "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/Xiaomi/OTA_lumi.plug.sacn03_0.0.0_0034_20211119_85CEC0.ota",
         "path": "images/Xiaomi/OTA_lumi.plug.sacn03_0.0.0_0034_20211119_85CEC0.ota"
+    },
+    {
+        "fileVersion": 10003,
+        "fileSize": 264682,
+        "manufacturerCode": 4098,
```
Koenkk, line 1415: Please also add the manufacturer names similar to https://github.com/Koenkk/zigbee-OTA/blob/5acb22ab5e123ff678ed007f91e44a83bbba1744/index.json#L753 (this prevents this update being loaded on other devices)

heinekmadsen:
not sure how i find that info similar to your reference
image

also.. the locks is more or less the same as this, but not all the Exposes are the same..... can i easily test if these Exposes work with the Wattle branded version?
https://www.zigbee2mqtt.io/devices/easyCodeTouch_v1.html
https://www.zigbee2mqtt.io/devices/HC-SLM-1.html (Wattle)

Koenkk:
I see this lock uses a different manufacturer naming mechanism. Then add a modelId instead: https://github.com/Koenkk/zigbee-OTA/blob/7cfc1503837b9c8666b007946e8a96c32a041949/index.json#L814 (HC-SLM-1).

> also.. the locks is more or less the same as this, but not all the Exposes are the same..... can i easily test if these Exposes work with the Wattle branded version?

If you add the modelId then this fw will only by loaded on the HC-SLM-1 (which is what we want since this fw has not been tested on other devices and could potentially brick it)

heinekmadsen:
Updated with modelId

Koenkk:
Thanks!

heinekmadsen:
@Koenkk Still struggeling with this Wattle lock and firmware... Turns out there's a newer version, and i have found that and will add it to this repo, but it turn out there's two versions of the fw.. one for a PA version and one for a P0 version.
I cant find any thing in the software/zigbee2mqtt where i can tell which version it is, but looking at the physical device i can tell which one to use.

How do we overcome that?
https://heimgard.com/support/firmware-filer-til-vare-produkter

Koenkk:
Can the PA and P0 version be distinguished based on the manufacturer name or zigbee model id? Then something like https://github.com/Koenkk/zigbee-OTA/blob/05b1b83ff3b2d2e6dcebb0f9cb9158e0844bdc7c/index.json#L753 can be done.

LSvense:
@heinekmadsen Did you manage to get your Wattle lock updated? I got the same lock (with PA version of the zigbee module), and was glad to see it added to the repo, but not sure what to do from here, the lock doesn't appear on the OTA page.

heinekmadsen:
> @heinekmadsen Did you manage to get your Wattle lock updated? I got the same lock (with PA version of the zigbee module), and was glad to see it added to the repo, but not sure what to do from here, the lock doesn't appear on the OTA page.

I did manage to get it updated, but i used zha as i had a spare sonoff zigbee bridge. I still haven't found any way to identify the hardware version :(

@Koenkk maybe we should remove the firmware from this repository so we don't accidently break any of the P0 devices

Koenkk:
@heinekmadsen removed it.

LSvense:
@heinekmadsen Thanks for the hint - I too made a temporary ZHA installation, thus enabling OTA from a local file and managed to get the lock updated. I can see that the lock now expose two new attributes (auto_relock_time and sound_volume) - do you know enough about zigbee-herdsman-converters and GitHub pull requests to include these attributes in what Z2M exposes? I unfortunately don't.

PR #79, opened 2022-01-02
Update Philips Hue smart plug firmware

davidjb:
Updates the firmware for the Hue smart plug to 1.93.6 (https://www.philips-hue.com/en-us/support/release-notes/accessories).

Commits for intermediate versions between the now and the latest version have been included for posterity - they might be useful for some reason later.


Koenkk:
Thanks!

PR #90, opened 2022-01-28
Add QBKG40LM and QBKG41LM ota files

zvldz:
Add lumi.switch.b1nc01(QBKG40LM) and lumi.switch.b2nc01(QBKG41LM) ota files


Koenkk:
thanks!

PR #97, opened 2022-02-25
Updated firmware for lumi.curtain.acn002 to version 1427

schwickster:
Latest firmware for lumi.curtain.acn002. Not sure if the fileVersion is correct. URL and fileSize should be correct.


Koenkk:
thanks!

PR #98, opened 2022-02-26
Fix lumi.curtain.acn002 update

TheJulianJES:
It seems like the ``add.js`` script wasn't used in https://github.com/Koenkk/zigbee-OTA/pull/97.

(Also, it looks like the ``add.js`` script adds a second entry and doesn't update the existing one in this case. (Something to do with ``modelId``?) Might be a similar issue with the ``updateall.js`` script and Hue's version stuff.)


Koenkk:
thanks!

PR #112, opened 2022-05-04
Updated Hue image

TheJulianJES:
Updated Hue image 100B-0118 PixelLum-EFR32MG21

(The update is not yet available through the Hue Bluetooth app)


Koenkk:
Thanks!

Issue #114, opened 2022-05-13
ERR_BUFFER_OUT_OF_BOUNDS


mdeweerd:
I've already applied this update and I wanted to add it to the list for others:

```shell
node scripts/add.js https://github.com/fairecasoimeme/Zlinky_TIC/releases/download/v5.0/ZLinky_router_v5.0.ota
```

However I get:
```raw
node:internal/errors:466
    ErrorCaptureStackTrace(err);
    ^

RangeError [ERR_BUFFER_OUT_OF_BOUNDS]: Attempt to access memory outside buffer bounds
    at new NodeError (node:internal/errors:377:5)
    at boundsError (node:internal/buffer:84:11)
    at Buffer.readUInt16LE (node:internal/buffer:243:5)
    at Object.parseImage (D:\mdeweerd\workspace\zigbee-OTA\lib\ota.js:17:34)
    at main (D:\mdeweerd\workspace\zigbee-OTA\scripts\add.js:65:24)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  code: 'ERR_BUFFER_OUT_OF_BOUNDS'
}
```

I tried `node scripts/add.js http://fds.dc1.philips.com/firmware/ZGB_100B_010D/1107323831/Sensor-ATmega_6.1.1.27575_0012.sbl-ota` which executed without any issue.

mdeweerd:
I get the same message with, but I do not know if zip files are supported:
```shell
node scripts\add.js https://files.danfoss.com/download/Heating/Ally/Danfoss%20Ally%20Radiator%20Thermostat%201.18%20OTA%20file%2001122021.zip
```

Also, when performing:
```shell
wget https://github.com/fairecasoimeme/Zlinky_TIC/releases/download/v5.0/ZLinky_router_v5.0.ota
node scripts/add.js ZLinky_router_v5.0.ota
```

I get

```raw
Added new entry ({"fileVersion":5,"fileSize":240478,"manufacturerCode":4151,"imageType":1,
"sha512":"692417e8863827f120cda063850aee65b7b6eb58109d419ff58dced1d7c9251617312368bda0f848
8aa26aeb7a8e3642339c7648af24cbff66b870401c9113da","url":"https://github.com/Koenkk/zigbee-
OTA/raw/master/images/Eurotronic/ZLinky_router_v5.0.ota","path":"images/Eurotronic/ZLinky_
router_v5.0.ota"})
```

[...]

PR #120, opened 2022-05-16
Add lixee (third party link)

mdeweerd:
Earlier push included a local path to ZLinky due to tests for #114 .
Corrected this by adding link and removing path (manually, add.js doesn not remove the path).


Koenkk:
Thanks!

PR #131, opened 2022-06-27
Added older Hue Adore image

TheJulianJES:
Untested
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/12954


Koenkk:
Awesome, thanks!

Tamas-Toth-ebola:
Hi!

As it seems that more people run into this problem, again and again, as I wrote below, would be nice to 'sharply' highlight the 'problem' in any documentations to avoid it in the future:

https://github.com/dresden-elektronik/deconz-ota-plugin/issues/10#issuecomment-1109482890 ... and the next 4-5 comments.

Anyway thanks for all your works in this repo!

PR #148, opened 2022-08-17
add csm-300z v2

nukusinji:
Add ShinaSystem CSM-300Z V2

Manufacturercode is same between Jennic and Shinasystem.

So I just add ota file in /images/

And I will update device converter, when I testing ota update by this PR.


Koenkk:
Thanks!

PR #149, opened 2022-08-18
fix path of csm-300zb

nukusinji:
fix path #148


Koenkk:
Please move the image file to the `Jennic` folder instead.

nukusinji:
> Please move the image file to the `Jennic` folder instead.

Okay.

Koenkk:
Thanks!

PR #154, opened 2022-08-30
Update ZLinky to V6

mdeweerd:


Koenkk:
Thanks!

ValentinAUCLERC:
It brokes my device... The manufacturer tells that it's an experimental version that shouldn't be used in production, please revert as soon as possible !

mdeweerd:
On french forum, users indicate they need to re-associate, but the device is not bricked as such.

https://forum.hacf.fr/t/zlinky-tic-ou-connecter-simplement-linky-en-zigbee/7112

Submitted PR: #156

ValentinAUCLERC:
The manufacturer clearly say here https://github.com/fairecasoimeme/Zlinky_TIC/releases that the v6 is experimental and shouldn't be used in production.

Now only my LQI is updated and I'm struggling to downgrade if possible

mdeweerd:
@ValentinAUCLERC In the mail I received and at the time that I checked the changes that were made to the code, the release did not mention that it was experimental and the release was not labeled as a pre-release.

PR #156, opened 2022-09-06
Z linky v6 -> v5

mdeweerd:
https://github.com/fairecasoimeme/Zlinky_TIC/releases/tag/v6.0 now lists this as a pre-release which was not the case earlier.

On french forum, users indicate they need to re-associate, but the device is not bricked as such.

https://forum.hacf.fr/t/zlinky-tic-ou-connecter-simplement-linky-en-zigbee/7112

Anyway, reverting this.


Koenkk:
There are no changes in this MR (https://github.com/Koenkk/zigbee-OTA/pull/156/files)

mdeweerd:
Ok, I rebased now.

PR #158, opened 2022-09-07
DMS-300ZB V2 Firmware

shinasys:
update


Koenkk:
An OTA file was removed but none has been added back. can you check this?

shinasys:
> An OTA file was removed but none has been added back. can you check this?
This OTA File is modified(Deleted and added.)

Koenkk:
I see, thanks!

PR #140, opened 2022-07-16
Update OTA file for Xiaomi `SP-EUC01`

Otnow:
```json
"releaseDate": 1642585695000,
"part": "zigbee",
"downloadUrl": "https://cdn.aqara.com/cdn/opencloud-product/mainland/product-firmware/prd/lumi.plug.maeu01/20211209165104_OTA_lumi.plug.maeu01_0.0.0_0041_20211206_0C22EC.ota",
"upgradeTakesTime": 0,
"updateLog": "1.Fix the offline problem of device",
"updateTime": 1642585695000,
"isNecessary": 0,
"fileSize": 273278,
"createTime": 1642585695000,
"firmwareMD5": "775822596abe1fc2820b59bbcb0c22ec",
"state": 3,
"isForce": false,
"firmwareVersion": "0.0.0_0041",
"imageType": "exe",
"firmwareCRC32": "B575A675"
```

PR #159, opened 2022-09-09
Return previous OTA file for Xiaomi `SP-EUC01`

Otnow:
Reverting to previous version `0.0.0_0032` because version [0.0.0_0041](https://github.com/Koenkk/zigbee-OTA/pull/140) is [causing problems](https://github.com/Koenkk/zigbee2mqtt/issues/13903).


TheJulianJES:
FYI, this is fixed in ZHA by just making the plug leave group 0: https://github.com/zigpy/zha-device-handlers/pull/1706

PR #160, opened 2022-09-10
add latest Danfoss Icon Zigbee image

nagydavid:
update to the latest Danfoss Icon Zigbee image


Koenkk:
Please also remove the existing one from `index.json` and the OTA file itself.

PR #161, opened 2022-09-16
Added new firmware for the WS-EUK03 (lumi.switch.n1aeu1)

octa22:
This PR adds OTA firmware for WS-EUK03 devices


Koenkk:
Thanks!

PR #171, opened 2022-10-22
New Firmware for DALI CONV Li

Danit2:


Koenkk:
Please add it as described on https://github.com/Koenkk/zigbee-OTA#adding-new-and-updating-existing-ota-files (this will add it to the `index.json`)

PR #176, opened 2022-11-03
Develco AQSZB-110 fw 4.0.1

sjorge:
@Koenkk I tested this on 1 of the 2 out dated AQSZB-110 with the local override.

With this content

```
[
    {
        "url": "ED%20-%20HA%20-%20VOC%20Sensor-SSIG%204.0.1.zigbee"
    },
    {
        "url": "ED%20-%20HA%20-%20VOC%20Sensor-SSIG_B%204.0.1.zigbee""
    }
]
```

But It seems they are the same with a different file version, so I think only one might be needed?

I have one more out dated device and I was planning on updating it when this gets merged (well and a zhc PR to add the ota tag to the device)

I got both of these in a single zip archive from the friend support email.

Both looked like this:
image

After updating 1/2 on the older fw it looked like this, it did take over 4h!
image

It wasn't clear from the logs which of the 2 it used unfortunately.


sjorge:
Wait I think this means it used the first one?

```
debug 2022-11-03 13:55:49: getNewImage for '0x0015bc0036000d10', meta {"fileVersion":262145,"url":"ED%20-%20HA%20-%20VOC%20Sensor-SSIG%204.0.1.zigbee"}
```

Should I remove the other one?

sjorge:
@Koenkk OK it's the non _B image, I tried again with only the _B image in the OTA neither the up-to-date or the out-of-date device want to use that one.

I'll do a zhc PR once this gets merged and I have tested with the device I kept aside.

Koenkk:
Keeping only one makes sense (assuming this device doesn't need to updated in two steps).

sjorge:
> Keeping only one makes sense (assuming this device doesn't need to updated in two steps).

Given neither the old or new fw seem to think the _B variant is an update, i’d say thats a safe assumption.

I suppose there could be an even older fw that might need it as an extra step?

sjorge:
My other one is updating now, if it succeeded in the morning, i’ll open the zhc PR. So far so good, just slow.

PR #210, opened 2022-12-14
Add Namron firmware

atus:
Adding fw for
Thermostat
Black/White 4/8 button switch
4-in-1 dimmer
Rele


Koenkk:
Can you also create a pr for https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/namron.js ?

atus:
Sure. I'll check it out.

atus:
Hmm, I guess there's some code needed to support OTA for Namron devices. I'm not familiar with this code but I can try if it's simple enough. If you can spell it out or if there's an equivalent PR, I can do it.
Comparing to Philips.js I see `const ota = require('../lib/ota');` in the header and `ota: ota.zigbeeOTA,` sprayed all over the place. And perhaps removing `ignore_genOta`? Is that it?

atus:
https://github.com/Koenkk/zigbee-herdsman-converters/compare/master...atus:zigbee-herdsman-converters:patch-1

Koenkk:
Thanks!

PR #221, opened 2022-12-29
Updated firmware for Datek HAN Adapter HW revision 1 to 0.7

TheStigh:
Updated firmware for Datek HAN Adapter HW revision 1 to 0.7


Koenkk:
Thanks!

PR #202, opened 2022-12-05
added ota files for microdim 07 dimmer/switch

Kilill:
ota files from ecodim.nl for "Eco-Dim.07/Eco-Dim.10" uploaded after permission from Gabbe van der Zwan

PR #226, opened 2023-01-02
New OTA file for eco-dim-07 with corrected manufacturer id

Kilill:
New OTA file  received from EcoDim
Should now have the correct manufacturer id.
Also added EcoDim as manufacturer to add.js


index.json
```diff
@@ -2363,5 +2363,14 @@
         "sha512": "b4d6e2ee31ad7a90ec0d87d3a00ca08a587ac5ca276484a2ffdc8ff94dcb6b63199f679e6659a2b5217aa4a6bf067c8bc10692ca541ddec354184939a673cb71",
         "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/Namron/4512719-Firmware-35.ota",
         "path": "images/Namron/4512719-Firmware-35.ota"
+    },
+    {
+        "fileVersion": 10,
```
erlendk, line 2368: @Kilill The fileVersion is still pretty long away from the one both you and I have (50947376). Don't know how that has happened, but it might mean that a force update will be required.

erlendk:
Maybe this PR should also revert the changes made in https://github.com/Koenkk/zigbee-OTA/pull/202 ?

Kilill:
Yes the #202 should be reverted as well as this one which has an incorrect version compared to the original
EcoDim has been poked to se if they can deliver an updated one...

Koenkk:
Is this ready for merge?

Kilill:
No, drop it, turn out the file has incorrect version info and some kind of corruption in the image

PR #242, opened 2023-01-25
Path correction for UHome OTA

AlexBolshov:
Fixing OTA Error 404 This was the wrong path so I fixed it. Please tell me, are these all the places where I need to change it?


Koenkk:
Looks complete, thanks!

PR #243, opened 2023-01-25
Update Ubisys range to new firmwares - ubisys Zigbee Stack 4.0.3

geirra:
New firmware for the D1 dimmer was released 25.01.23

Changelog here:
https://www.ubisys.de/en/support/firmware/changelog-d1/

### DIFF
--- a/index.json
+++ b/index.json
@@ -1664,12 +1664,12 @@
         "url": "https://www.ubisys.de/wp-content/uploads/10F2-7B39-0000-0004-02000300-spo-fmi4.ota.zigbee"
     },
     {
-        "fileVersion": 33620736,
-        "fileSize": 145662,
+        "fileVersion": 34604035,
+        "fileSize": 149246,
         "manufacturerCode": 4338,
         "imageType": 31537,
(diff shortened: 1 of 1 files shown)


geirra:
Added all the other firmwares also.
ref: https://www.ubisys.de/en/support/firmware/

Koenkk:
@sjorge did you already check these?

sjorge:
Yes, but I'm confused why they are here too... AFAIK ota/ubisys.js fetches these directly from the ubisys website!

Koenkk:
I see, we also don't use this (Zigbee-OTA) repo for Ubisys devices. @geirra what is the reason for this pr?

geirra:
Error from my side, didn`t know that these where pulled for ubisys.js in the herdsman-converts.

I'll close this @Koenkk

sjorge:
@Koenkk I guess we should remove the old once as they are not being used?

geirra:
the older entries in index.js is the reason why i tought the firmware should be provided this way.
So might be a good idea

Koenkk:
Removed them!

Issue #265, opened 2023-02-25
Can not find OTA image for Danfoss Ally TRV


gedemagt:
Hi,

I recently bought a Danfoss Ally TRV, which is nicely supported by Z2M, and there are OTA updates in this repo.
However, when I try to check for updates, it says that it could not find any images for imageType 288
![image](https://user-images.githubusercontent.com/2789746/221370868-8654dac4-0042-4694-adb9-303d45b79ad6.png)

I can see that in you index.json in this repo, the imageType for Danfoss is also 256. 

Can you help me with why this doesn't recognize the correct version and type? My device looks like this:
![image](https://user-images.githubusercontent.com/2789746/221370947-ad6effe1-18fe-426c-9a21-3030916da2a1.png)

lasse-aagren:
I have same issue with newly bought ally's

red-island:
As I do no permission to create a branch/pull request here the solution for the coders:

This is the new model from Danfoss Ally => there is a new image id 0x0120
Hence, index.json needs to get a new entry.

Copy record with "manufacturerCode": 4678, "imageType": 256
and add it with new **"imageType": 288**

![image](https://user-images.githubusercontent.com/114957153/222952220-3cb8d0a7-a69a-4221-9969-e9f2f6d661fe.png)

gedemagt:
@Koenkk I created a PR for the suggestion by red-island. Please review: https://github.com/Koenkk/zigbee-OTA/pull/274

gedemagt:
Hmm, I tried to download the latest and upload it locally, but I keep getting an "imageType mismatch" error. I downloaded the firmware from here: https://www.danfoss.com/en/products/dhs/smart-heating/smart-heating/danfoss-ally/danfoss-ally-support/#tab-software and the .pdf that comes along clearly states that is should be the new imageType of 288 (0x120). 

Is it z2m or the device that responds check the imageType?

hotchkj:
I'm unconvinced Danfoss have actually set this data up correctly. I'm in the same situation, and the debug logs for the OTA file indicate:

```
[...]

PR #274, opened 2023-03-05
Updated index.json with new danfoss ally entry.

gedemagt:
According to [red-island](https://github.com/Koenkk/zigbee-OTA/issues/265#issuecomment-1455035466) there is a new image id for Danfoss Allys. So i created a new entry.

### DIFF
--- a/index.json
+++ b/index.json
@@ -958,6 +958,15 @@
         "path": "images/Danfoss/1246-0100-01200120.0002_(2B60A723).ota"
     },
     {
+        "fileVersion": 276,
+        "fileSize": 385719,
+        "manufacturerCode": 4678,
+        "imageType": 288,
+        "sha512": "6b89fd90146747fde88e4896a3cb7b8600ee733679d234cb0288632c820264fcb8e5c25a79a04a4faa2114742a6f3fd37b107be6c60e927eea253fe6d8744d6a",
+        "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/Danfoss/1246-0100-01200120.0002_%282B60A723%29.ota",
+        "path": "images/Danfoss/1246-0100-01200120.0002_(2B60A723).ota"
+    },
+    {
         "fileVersion": 5,
         "fileSize": 291622,
         "manufacturerCode": 4687,
--- a/index.json
+++ b/index.json
@@ -2377,3 +2377,4 @@
         "path": "images/Xiaomi/20230110152619_OTA_lumi.magnet.agl02_0.0.0_0030_20230109_D7EBD0.ota"
     }
 ]
+


Koenkk:
Did you verify this change locally? Because now the imageType in index.json will mismatch the imageType from the `.ota` file

gedemagt:
No you are right - it conflicts on the image from the .ota file. I will try to contact Danfoss to hear if they have an available .ota file available.

gedemagt:
> Did you verify this change locally? Because now the imageType in index.json will mismatch the imageType from the `.ota` file

How do you see what the imageType in the .ota file is?

Koenkk:
You can see it in the `index.json` entry after running the add script: https://github.com/Koenkk/zigbee-OTA#adding-new-and-updating-existing-ota-files

gecko17:
The requested change here just adds two different entries, but point to the same image....

@Koenkk 
I think there should be two entries, one with the old imageType 256 pointing to the old image and one with the new imageType 288 pointing to the new image (which also has this type inside the image). And of course, we need to add the old image (1.18) again to the directory.

```
{
        "fileVersion": 276,
        "fileSize": 385719,
        "manufacturerCode": 4678,
        "imageType": 288,
        "sha512": "6b89fd90146747fde88e4896a3cb7b8600ee733679d234cb0288632c820264fcb8e5c25a79a04a4faa2114742a6f3fd37b107be6c60e927eea253fe6d8744d6a",
        "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/Danfoss/1246-0100-01200120.0002_%282B60A723%29.ota",
        "path": "images/Danfoss/1246-0100-01200120.0002_(2B60A723).ota"
},
{
        "fileVersion": 274,
        "fileSize": 379806,
        "manufacturerCode": 4678,
        "imageType": 256,
        "sha512": "873effaa113e1d07643fbf25f6e26ee4b23f8deb7d006cce7fc490fd8ad7a5754de4c352b6f3d5e172022a34ea2699f5d1d8fda90f4348b25d5feaa169e38c85",
        "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/Danfoss/1246-0100-01180118.0002_%28F5BFD8B8%29.ota",
        "path": "images/Danfoss/1246-0100-01180118.0002_(F5BFD8B8).ota"
},
```

If I'm correct, the current entry in the master is wrong, because it says imageType 256, but the imageType inside the .ota file is 288?

Koenkk:
@gecko17 could you provide me `Danfoss/1246-0100-01180118.0002_%28F5BFD8B8%29.ota`?

gecko17:
> @gecko17 could you provide me `Danfoss/1246-0100-01180118.0002_%28F5BFD8B8%29.ota`?

@Koenkk it is in the git history available at https://github.com/Koenkk/zigbee-OTA/blob/7f3800dfd5995bbb9cb1ec62ecccf879e3b6a9c2/images/Danfoss/1246-0100-01180118.0002_(F5BFD8B8).ota

So you can just check it out and restore it with `git`.

I have it somewhere on my old notebook, but would be easier to use the file from the history.

red-island:
Alternative source of v1.18: https://files.danfoss.com/download/Heating/Ally/Danfoss%20Ally%20Radiator%20Thermostat%201.18%20OTA%20file%2001122021.zip

gedemagt:
As far as I can see, none of these files have 288 as their imageType within the .ota file. I have tried the following:

Version 1.18: 
Updated existing entry ({"fileVersion":274,"fileSize":379806,"manufacturerCode":4678,"imageType":256,"sha512":"873effaa113e1d07643fbf25f6e26ee4b23f8deb7d006cce7fc490fd8ad7a5754de4c352b6f3d5e172022a34ea2699f5d1d8fda90f4348b25d5feaa169e38c85","url":"https://github.com/Koenkk/zigbee-OTA/raw/master/images/Danfoss/1246-0100-01180118.0002_%28F5BFD8B8%29.ota","path":"images/Danfoss/1246-0100-01180118.0002_(F5BFD8B8).ota"})

Version 1.20:
Updated existing entry ({"fileVersion":276,"fileSize":385719,"manufacturerCode":4678,"imageType":256,"sha512":"6b89fd90146747fde88e4896a3cb7b8600ee733679d234cb0288632c820264fcb8e5c25a79a04a4faa2114742a6f3fd37b107be6c60e927eea253fe6d8744d6a","url":"https://github.com/Koenkk/zigbee-OTA/raw/master/images/Danfoss/1246-0100-01200120.0002_%282B60A723%29.ota","path":"images/Danfoss/1246-0100-01200120.0002_(2B60A723).ota"})

red-island:
@gedemagt 
Hi Jesper, 
when checking the Danfoss Ally webpage https://www.danfoss.com/en/products/dhs/smart-heating/smart-heating/danfoss-ally/danfoss-ally-support/#tab-software - download v1.20, the embedded pdf state that network identifier changes from image type id 0x0100 to new id 0x0120 (to me this is the change from 256 to 288). As also the model changes I assume the change targets the new Ally models eTRV0103 only.

Anyhow, I could install v1.20 on the old eTRV0100 and the only defect I have is the silly version (see the defect I raised).
Is maybe the OTA targetting 256 and 288 in a single ota image?

I am not the export on all this, so just summarizing my point of view.

gedemagt:
@red-island 
I see the same behavior, but when I try to update my eTRV0103 with the new v1.20 via Z2M OTA it complains that the imageType dosen't match - which I suppose is because the imageType in the header of the .ota file is still 256. So even if it does target the new ones, I am somehow not allowed to use it - not sure if it is due to Z2M or the TRV's. 

I tried to ask Danfoss why the header in the .ota file is 256 and not 288, but they didn't respond with anything useful. 

Do you know if it is somehow possible to "override" the check?

Koenkk:
Based on this, it seems we should add the 1.20 firmware. When updating to this firmware the imageType changes to 288 so you could do another update (??). Such two-step update approach is not uncommon, Philips also does this.

gedemagt:
I still have the issue that I can not update to 1.20 or 1.18 since they both have the old imageType in the .ota file. Any idea how to solve that?

red-island:
@gedemagt 
Hi Jeasper, to me it would make sense if your brand new eTRV0103 came with 1.20 imageType 288. Fron rhe pdf it sounds that they had to do 1.20 update for the old devices to have single version for both going forward.
What firmware is on your new device?

gedemagt:
I totally agree, and I am quite puzzled that it has image version 20 only:
image

red-island:
The 1.18 is from 2021
https://assets.danfoss.com/documents/193613/AM375549618098en-000102.pdf

So yours is newer.
Maybe 0.20 and 1.20 is the same = v20

PR #275, opened 2023-03-07
Update OTA file for Xiaomi `SRTS-A01`

leroyloren:
Version: 0.0.0_1030 2023-03-07

What's New:
1. Add 3 new schedules.
2. Added temperature curve in app.
3. Notification of Valve test is off by default.
4. Fix other known issues.


Koenkk:
Please update the existing entry: https://github.com/Koenkk/zigbee-OTA/blob/f1f6308d43c4d0ded548426948eecf8e10288706/index.json#L2022

leroyloren:
Ok, done.

Koenkk:
Thanks!

TheJulianJES:
@leroyloren Do you have this device and did you test the update?

With both ZHA and Z2M (and even when not connected to any network), I was unable to go from "calibration"/setup (F11) in to "normal operating" mode (after temporarily taking out the batteries).
It would always go from F11 -> three button presses -> F12 -> couple of seconds -> F11.

I've downgraded back to 0930 and everything seems to work as expected again.

leroyloren:
It works without a problem, I don't notice any unusual behavior.

PR #278, opened 2023-03-09
Add firmware version 0043 for Aqara ``SP-EUC01``

DerDreschner:
Response from web service:
```
"deviceModel": "lumi.plug.maeu01",
"deviceName": "Smarte Steckdose (EU-Standard)",
"firmwareType": "zigbee",
"firmwareVersion": "0.0.0_0041",
"upgradeFirmware": {
    "createTime": 1676605625000,
    "downloadUrl": "https://cdn.aqara.com/cdn/opencloud-product/mainland/product-firmware/prd/lumi.plug.maeu01/20230129165424_OTA_lumi.plug.maeu01_0.0.0_0043_20230113_7E5AB5.ota",
    "fileSize": 277486,0041
    "firmwareCRC32": "72C4FC24",
    "firmwareMD5": "e2af34c841d425dbe80cb6e4517e5ab5",
    "firmwareVersion": "0.0.0_0043",
    "imageType": "exe",
    "isForce": false,
    "isNecessary": 0,
    "part": "zigbee",
    "releaseDate": 1677553080000,
    "state": 3,
    "updateLog": "Function optimization.",
    "updateTime": 1677553080000,
    "upgradeTakesTime": 0
}
```

I'm not sure if this update should be rolled out, as the version 0041 have been reverted due to problems (see #159). Although, my plugs came with version ``0.0.0_0041`` from the factory and work perfectly fine with Zigbee2MQTT.

Also, is it desired to delete old firmware versions inside the ``images`` folder as well es the index.json? I didn't do that here as I wasn't sure. Can do that if it's considered good practice.


Koenkk:
Please update the existing entry (https://github.com/Koenkk/zigbee-OTA/blob/dbf4b37bcc4bb839006c07315e5194b1cee38031/index.json#L1064) and remove the old image.

DerDreschner:
> Please update the existing entry and remove the old image.

Thanks for the quick response! Implemented the changes.

TheJulianJES:
This firmware is broken like 41 and re-introduces the random toggling when Zigbee group 0 is used anywhere. There are multiple issues on this.

Koenkk:
@TheJulianJES thanks for letting us know!

DerDreschner:
Sad to see this isn't fixed yet. Used a custom `index.json` file to flash my plugs with this version as I'm not affected by this bug due to a custom Zigbee PAN. I close this PR then.

@Koenkk: FYI, this firmware introduces two new entities.

```
    "consumer_connected": false,
    "consumption": 0.57,
```

`consumption` is the same value as `energy` while `customer_connected` states if any plug is inserted. Is is possible to expose entities if they are available/the firmware version is above a specific value? If yes, which file in Zigbee2MQTT I have to edit to expose at least the plug detection entity?

TheJulianJES:
Both entities are already available with the earliest firmware known (at least with ZHA).

DerDreschner:
Ohh, really? Interesting. I'm not that far into how everything works (like any filters e.g.), but I think I didn't had these entities under `state` before the update. Don't have any plugs with the old 0041 firmware to confirm. That's why I thought they would be new. :smile:

TheJulianJES:
It's available with v32 (non-buggy firmware) here

Kars-de-Jong:
> Sad to see this isn't fixed yet. Used a custom `index.json` file to flash my plugs with this version as I'm not affected by this bug due to a custom Zigbee PAN. I close this PR then.

@DerDreschner What does that mean, "a custom Zigbee PAN"? Can I configure that too to work around this issue?

PR #284, opened 2023-03-13
Added manufacturer EcoDim + OTA file

sandert98:
Eco-Dim.10 OTA file version 1.01

I'm not 100% sure if this is correct and fixes my issue. 
I requested the OTA file for the EcoDim 10 at the manufacturer, and I got this file. There is already an EcoDim OTA file with imagetype 44 and version 10, but it has manufacturer Tuya (Rows 2118 / 2125). 
The error that is thrown in my log consists of imageType 60, as described in https://github.com/Koenkk/zigbee2mqtt/issues/16999


Koenkk:
Thanks!

PR #285, opened 2023-03-13
OTA File for IKEA E2123 (new Symfonisk Remote)

ilueckel:
This merge request contains the OTA file sniffed by the guys over at https://github.com/zigpy/zigpy/discussions/660.

The remote only updates with the new IKEA Dirigera Hub, which does not use the index file of the old Tradfri Gateway. So there seems no way to fetch the new updates via URL.


Koenkk:
Thanks!

PR #294, opened 2023-04-18
update namron thermostat (4512737/4512738)

ruant:
This firmware update resolves the issue with this device using random zigbee devices on the network to control it's occupancy mode (aka. `away_mode`). This made it randomly activate the `away_mode` feature on the device.

I've been running this version for a few days now and I have now issues with it.

BUT! There is a code change in `zigbee-herdsman-converters` that is needed before this device will configure itself correctly after the firmware upgrade. Due to the occupancy stuff being removed from the device now.

Converter PR:


ruant:
@Koenkk 
Looks like the `add` node script created duplicate entries for these devices?..

4512737:
Existed: https://github.com/Koenkk/zigbee-OTA/blob/7e8a7ef518b982106f7ee95a3f2420a18ac86bdd/index.json#L2165-L2172
Newly added: https://github.com/Koenkk/zigbee-OTA/blob/7e8a7ef518b982106f7ee95a3f2420a18ac86bdd/index.json#L2614-L2620


4512738:
Existed: https://github.com/Koenkk/zigbee-OTA/blob/7e8a7ef518b982106f7ee95a3f2420a18ac86bdd/index.json#L2215-L2222
Newly added: https://github.com/Koenkk/zigbee-OTA/blob/7e8a7ef518b982106f7ee95a3f2420a18ac86bdd/index.json#L2623-L2629

Koenkk:
That's because the existing entry has a `modelId` and in that case it is not updated automatically, can you update it manually?

Koenkk:
Thanks!

PR #298, opened 2023-04-28
Fix Third Reality URLs from firmwares to firmware

digiblur:
@tube0013 mentioned in discord that the URLs were "firmware" and not "firmwares"  - the new links look to be working.


Koenkk:
Thanks!

PR #301, opened 2023-05-04
Added wb-msw3 ota fw

aadegtyarev:


Koenkk:
Thanks, could you also make a pr for https://github.com/Koenkk/zigbee-herdsman-converters?

aadegtyarev:
> Thanks, could you also make a pr for https://github.com/Koenkk/zigbee-herdsman-converters?

What should be sent there? There is a converter for this device: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices/sprut.js

Koenkk:
`ota: ota.zigbeeOTA,` is already there so that is fine, thanks!

PR #303, opened 2023-05-11
Updated Hue images

TheJulianJES:
This updates 7 Hue images. Firmware seems to be: 1.104.2
For the changelog, see: https://www.philips-hue.com/en-us/support/release-notes/lamps


Koenkk:
Thanks!

PR #304, opened 2023-05-15
Added fw v55 for Legrand 067776, v78 for Legrand 412171

benjR:
Used wireshark to get the ota file, tested on other devices and ok


Koenkk:
Thanks!

benjR:
@Koenkk probably needed to change the ota parameter for those in the other project, will do a PR when I have time

PR #306, opened 2023-05-16
Bosch ```BTH-RA```: Add firmware 3.05.05 (may 2023 update)

DerDreschner:
Today, Bosch released a new firmware version for their smart thermostat 2. I sniffed and extracted the firmware and successfully updated another thermostat over Z2M with it.

Changelog can be found under https://www.bosch-smarthome.com/uk/en/release-notes/bosch-smart-home-system/2023-05-16.html and are copied here for completeness:
- Stability improvements during the update process
- Improvements to the heating algorithm with the aim of reducing temperature overshoot
- Faster heating phase with improved valve percentage calculation and more frequent valve openings when heating is required
- Prevention of unnecessary valve movements
- Correction of the problem relating to resynchronisation with the room thermostat II
- Various other minor bug fixes


Koenkk:
Thanks!

PR #308, opened 2023-05-19
Updated Hue image

TheJulianJES:
Seems to be a follow-up to:
- https://github.com/Koenkk/zigbee-OTA/pull/303

So also likely 1.104.2 for hardware version 100B-11D
Release notes: https://www.philips-hue.com/en-us/support/release-notes/lamps


Koenkk:
Thanks!

PR #311, opened 2023-05-24
feat(develco): add ota firmware for window sensor WISZB-120

ultrabug:
I plan on adding ota update capability for this device on the herdsman-converter project


Koenkk:
Thanks!

PR #315, opened 2023-05-26
sonoff zbminil2 ota image add

liangjia2019:
fix the bug of switching external switch types even with three consecutive keystrokes during network access


Koenkk:
Thanks!

PR #317, opened 2023-05-31
Add Ubisys OTA firmwares

WhistleMaster:
Add OTA firmwares for:
- C4
- D1
- J1
- R0
- S2
- S1-R


Koenkk:
There is no need to add them to this repo, they are retrieved from the Ubisys repo:

- https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/ota/ubisys.js
- http://fwu.ubisys.de/smarthome/OTA/release/index

WhistleMaster:
Thanks for the reply ! I didn't know that the ubisys firmwares were downloaded from their repo. 

However, the strange thing is that when I click on "Check for news updates" from the "OTA" tab for my Ubisys devices, I get the error "Firmware on 'XXX' is newer than latest firmware online.".

Koenkk:
Where did you get these OTA files from, from the same site?

WhistleMaster:
I've got them from ubisys web site: https://www.ubisys.de/en/support/firmware/

Koenkk:
Seems to be the same OTA files as listed on http://fwu.ubisys.de/smarthome/OTA/release/index (which is used for checking ATM)

WhistleMaster:
It seems so indeed, but I don't get why I have the error "Firmware on 'XXX' is newer than latest firmware online." when I click on "Check for news updates" from the "OTA" tab. 

Do you have any idea why ?

Koenkk:
That means the firmware on your device is newer than the firmware provided on the Ubisys site (which can happen if they did not publish the fw yet but provided it on the device delivered to you)

WhistleMaster:
The problem is that the firmware installed on the devices are not newer than the one available. The devices have OTA file version 02000300 and the one provided on the Ubisys web site is OTA file version 02120404.

Is there a way I can debut that or manually install the firmware ?

WhistleMaster:
Sure ! The only problem I have is for the S1, it seems to have two different firmware, based on the hardware version, and when I try to add the first one with the node script and then the second, it just replaces the hardware revision < 2 firmware by the one for the hardware revision >= 2.

Koenkk:
Reverted because of https://github.com/Koenkk/zigbee-OTA/issues/329

PR #339, opened 2023-08-01
Add Ikea Tradfri Shortcut E1812 2.3.015

peca89:
Add old firmware which removes double press function, but hence improves responsiveness of single-press function because it does not wait for second press.


peca89:
Oh, sorry, didn't realize it was different process...Any way of downgrading tradfri firmware then?

Koenkk:
You can try with the OTA index override: https://www.zigbee2mqtt.io/guide/configuration/ota-device-updates.html

PR #340, opened 2023-08-02
Update index.json

crispy78:
Added IKEA devices listed on http://fw.ota.homesmart.ikea.net/feed/version_info.json (excluding http://fw.ota.homesmart.ikea.net/global/GW1.0/01.20.065/bin/10032198-2.2-TRADFRI-gateway-1.20.65.p.elf.sig.ota.signed because of errors).

### DIFF
--- a/index.json
+++ b/index.json
@@ -2371,8 +2371,8 @@
         "fileSize": 305574,
         "manufacturerCode": 4476,
         "imageType": 4366,
-        "sha512": "771af218764fa9ec4a6f5fcda73c46afc3ac478e81cbf309e9356b093c6305db4d7d257d17adbcc4b91976171443e85a785cd4025ca160e5e716f43b3369d7e0",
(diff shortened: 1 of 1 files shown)


Koenkk:
There is no need to add these firmwares, they are already provided via `ota.tradfri` (e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/25a2a792ca80af3eadc0bc51c31999e48fb003c8/src/devices/ikea.js#L642)

crispy78:
The underlaying issue is that my Styrbar won't update from 1.0.024 to the most recent version. I've tried several methods, found Github and elsewhere, but due to my limited knowledge of running the Zigbee2MQTT Docker in combination with Home Assistant I haven't been able to update those devices.

crispy78:
Ikea is included in the software.

PR #342, opened 2023-08-04
Latest version from Gledopto GL-D-007P

BradleyFord:
They didnt provide me with a specific version number, other than saying that this is the latest version


Koenkk:
Thanks!

PR #351, opened 2023-09-02
Updated Hue image

TheJulianJES:
Likely software version 1.107.1 for "Philips Hue Festavia string lights with hardware platform version 100b-123"
Release notes: https://www.philips-hue.com/en-us/support/release-notes/lamps


Koenkk:
Thanks!

PR #355, opened 2023-09-11
Legrand New OTA

SilentT-FR:
Add all of new firmware from https://developer.legrand.com/production-firmware-download/


Koenkk:
Thanks!

PR #379, opened 2023-10-26
Update SIN-4-1-20

AlexisPolegato:
Update SIN-4-1-20 FW 1 to V2.31.0
Add SIN-4-1-20 FW 2 V1.1.1


Koenkk:
Please use the [add](https://github.com/Koenkk/zigbee-OTA#adding-new-and-updating-existing-ota-files) script when adding OTA files, this makes sure they get added to the `index.json`.

Koenkk:
I see the firmwares are for the same device, why did you upload both firmwares? (only 1 of them will be used)

AlexisPolegato:
The device has 2 different firmwares ; both can be updated.

Koenkk:
I see, just noticed that the `imageType` is different, thanks!

PR #392, opened 2023-11-06
Updated Hue images

TheJulianJES:
Releases for hardware versions: 100b-10c, 100b-117, 100b-11f
Release for 100b-10c is likely 1.108.7 which is in the [release notes](https://www.philips-hue.com/en-us/support/release-notes/lamps). The others do not appear to be in the release notes.

### DIFF
--- a/index.json
+++ b/index.json
@@ -160,12 +160,12 @@
     },
     {
         "minFileVersion": 16783874,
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #396, opened 2023-11-09
added Insta blinds actor firmware

runningman84:


runningman84:
I have added both versions add.js does not seem to be able to handle two files... only the latest entry stays there...

Koenkk:
- Why did you add 2 firmwares?
- Can you also make a PR to add `ota: ota.zigbeeOTA` for the relevant devices [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/insta.ts)?

runningman84:
I got two firmware version and I have a problem with one specific device. My idea was that maybe the old version might work better. Is it even possible to use zigebee2mqtt to downgrade a given firmware?

My problem is that I have two devices one works as designed the other does not report any status updates. The device (door) which is working fine shows an old firmware version in zigbee2mqtt but a new version in their app (how is this even possible?). The device (window) which does not report any status update shows the new version in their app and in zigbee2mqtt. This is very confusing...
[window.txt](https://github.com/Koenkk/zigbee-OTA/files/13328549/window.txt)
[door.txt](https://github.com/Koenkk/zigbee-OTA/files/13328550/door.txt)


I have added the device in my second PR https://github.com/Koenkk/zigbee-herdsman-converters/pull/6456

Koenkk:
- See the [docs](https://www.zigbee2mqtt.io/guide/usage/ota_updates.html#local-ota-index-and-firmware-files) on how to force a specific firmware
- Only 1 firmware should be added to this repo, I propose to pick the latest.

runningman84:
Okay I have deleted the older firmware. My config file looks like this:

Koenkk:
The device definition also needs to have `ota: ota.zigbeeOta` indeed (but that should be the case now in the dev branch) (https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html)

runningman84:
okay is there anything or can this be merged?

Koenkk:
thanks!

PR #406, opened 2023-12-02
New OTA Legrand

SilentT-FR:
I have multiple note about dev.legrand.com

Certain firmware cannot be applied if the device does not have a minimum version
I don't believe that Z2M and the architecture of the json file not permit this to be handled

I still left the files in _old, however there could be a problem if we use the updateall script, (order of listing of _old files after the recent ones (not tested)).

Perhaps adding a README to the folder would be relevant.

@Koenkk 

```
To be used if current NLF firmware version is lower than 28

To be used if current NLL firmware version is lower than 7

To be used if current NLV firmware version is lower than 21
```


Koenkk:
You can set the min/max file version, example: https://github.com/Koenkk/zigbee-OTA/blob/b6bf59ae3ffc4a6553390bc6a5482ffbe389e20c/index.json#L82

PR #411, opened 2023-12-11
Updated Legrand Firmware

grostim:
All firmware updated from https://developer.legrand.com/production-firmware-download/ as of latest update dated 2023, November 27.

image


Koenkk:
Thanks!

PR #414, opened 2023-12-17
Legrand MinVersion and remove old version

SilentT-FR:
the continue of : #406 but fastened by #411 @grostim be careful in your last PR you have mentionned the image of legrand

> All firmware updated from https://developer.legrand.com/production-firmware-download/ as of latest update dated 2023, November 27.
> 
> image

But in red you see the legrand firmware need a path to upgrade on very old firmware.

I cant find the version mentioned by Legrand so i use the last version possible, for there old firmwares, in "minversion" of last firmware

and i remove old firmware


SilentT-FR:
Thanks @Koenkk for the tips of minversion

Koenkk:
Thank you too!

PR #417, opened 2023-12-26
Fix Legrand NLV/NLL version detection

FabianMangold:
Add maxFileVersion to older NLV and NLL firmwares in order to fix the detection of the 'latest_version' available.


Koenkk:
Thanks!

PR #423, opened 2024-01-06
Update index.json to support _TZE200_kds0pmmv variant of TV02 thermostat from Moes

jmthill:
Add _TZE200_kds0pmmv variant of TV02 from Moes.
I tested it on 2 devices i own.


Koenkk:
Thanks!

PR #424, opened 2024-01-07
fix: Legrand NLV / NLL / NLP improvements

FabianMangold:
* Patched Legrand NLV / NLL / NLP binaries
* Fixed image detection when updating from older fw versions


Koenkk:
Thanks!

PR #429, opened 2024-01-15
Update EcoDim .07 Zigbee Pro Firmware Reference

Veldkornet:
The EcoDim .07 Zigbee Pro uses an existing firmware, but with a different imageType
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/20678


Koenkk:
Thanks!

PR #438, opened 2024-02-01
Add firmware 3.5 for Namron panel heaters

geirra:
Changes:

- Removed compensation table for setpoint
- Fix metering problems
- Zigbee stability improvements.

--------------

We need to update the device config also, as its config is missing OTA.

@Koenkk How is the update files matched against the device, for the panel heater its a single definition in z2m, but each device has it´s own firmware.
https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/namron.ts#L655


Koenkk:
It's matched on the combination of `manufacturerCode` and `imageType` which are send by the device on OTA update request

PR #441, opened 2024-02-05
Add custom firmware from pvvx for Xiaomi LYWSD03MMC

kwiateusz:
Resository the file came from: https://github.com/pvvx/ZigbeeTLc


Koenkk:
@devbis @krzysztof-kwitt asking because of https://github.com/Koenkk/zigbee-herdsman-converters/pull/7015, can I merge this?

devbis:
I don't have any objections. Though, this fork in it's current state it will not be supported by converters due to a different structure of clusters and the exactly same model/manufacturer.

kwiateusz:
If it will not be supported then there is no point in adding, thanks for the information

devbis:
@kwiateusz  It will be supported but without changes in pvvx's firmware https://github.com/pvvx/ZigbeeTLc/pull/52 only basic functionality (temperature + humidity) will be available with z2m. Either custom convertor will be required.

pbranly:
Hi
Could you explain the role of those additional settings ?
In z2m there are 2 ways to change temperature and humidity calibration. Which one is used ? 
In addition there are also settings for confort min/max temp and humidity. Same question : what is the role of those new settings 
Thanks 
Phil

devbis:
@pbranly Additional settings affect display values as well, while z2m internal calibration only modifies data in z2m and mqtt.
Comfort parameters is a configuration for happy/sad face on the display.

kwiateusz:
I've pushed a new version which shouldn't interfere with devbis converter

Koenkk:
@devbis @krzysztof-kwitt  can you do a final check?

devbis:
LGTM

Koenkk:
Thanks!

PR #425, opened 2024-01-08
Update OTA files for Xiaomi `QBKG30LM`, `QBKG31LM`, `QBKG32LM`, `ZNLDP13LM`, `TDL01LM` and `RTCZCGQ11LM`

Otnow:
**QBKG30LM**:
```json
"releaseDate": 1685328471000,
"part": "zigbee",
"downloadUrl": "https://cdn.aqara.com/cdn/opencloud-product/mainland/product-firmware/prd/lumi.switch.n1acn1/20230414104812_OTA_lumi.switch.n1acn1_0.0.0_211554_20230414_79A077.ota",
"upgradeTakesTime": 0,
"sourceFirmwareVersion": "0.0.0_xx54",
"updateLog": "Optimize the user experience\n",
"updateTime": 1685328471000,
"isNecessary": 0,
"fileSize": 325136,
"createTime": 1681799777000,
"firmwareMD5": "68a6c71e79f9f38194096713b779a077",
"state": 3,
"isForce": false,
"firmwareVersion": "0.0.0_1554",
"imageType": "exe",
"firmwareCRC32": "38F16727"
```
---
**QBKG31LM**:
```json
"releaseDate": 1685328354000,
"part": "zigbee",
"downloadUrl": "https://cdn.aqara.com/cdn/opencloud-product/mainland/product-firmware/prd/lumi.switch.n2acn1/20230414104625_OTA_lumi.switch.n2acn1_0.0.0_211554_20230414_3CB07D.ota",
"upgradeTakesTime": 0,
"sourceFirmwareVersion": "0.0.0_xx54",
"updateLog": "Optimize the user experience",
"updateTime": 1685328354000,
"isNecessary": 0,
"fileSize": 327424,
"createTime": 1681799865000,
"firmwareMD5": "f46ab574b17fad335cbb05c9243cb07d",
"state": 3,
"isForce": false,
"firmwareVersion": "0.0.0_1554",
"imageType": "exe",
"firmwareCRC32": "D8F8C455"
```
---
**QBKG32LM**:
```json
"releaseDate": 1685328276000,
"part": "zigbee",
"downloadUrl": "https://cdn.aqara.com/cdn/opencloud-product/mainland/product-firmware/prd/lumi.switch.n3acn1/20230414104924_OTA_lumi.switch.n3acn1_0.0.0_211554_20230414_0C7505.ota",
"upgradeTakesTime": 0,
"sourceFirmwareVersion": "0.0.0_xx54",
[...]

PR #443, opened 2024-02-18
Return previous OTA file for Aqara `RTCZCGQ11LM`

Otnow:
Reverting to previous version `0.0.0_0054` because version [0.0.0_0058](https://github.com/Koenkk/zigbee-OTA/pull/425) is [causing problems](https://github.com/Koenkk/zigbee-OTA/pull/425#issuecomment-1903533474) and it is [not possible to downgrade](https://github.com/Koenkk/zigbee-OTA/pull/425#issuecomment-1947407615) from it.


Koenkk:
Thanks!

PR #444, opened 2024-02-18
Add firmware 2.4.0 for Ubisys S1-R Series 2

frimtec:
Release notes: https://www.ubisys.de/unterstuetzung/support-firmware/change-logs-s1-r-series-2/


Koenkk:
This firmware is also available via [Ubisys directly](http://fwu.ubisys.de/smarthome/OTA/release/index), which is what z2m already uses. So this is not needed.

frimtec:
Thanks for the feedback. 
I was not aware that other sources are used as well.
I personally missing the 2.4.0 update for my devices, therefore I thought it is just missing in here.

PR #447, opened 2024-02-21
DIY lywsd03mmc OTA with changed manufacturer code

devbis:
During the last OTA I changed manufacturer code to custom to avoid struggling for image type numbers for updates. 
My suggestion is to put multiple DIY manufacturers to the same folder, I doubt there will be many of them. 
I also considered using `DIY_devbis` as a folder, but decided that it is redundant here.

Firmware here is the same as in Telink folder with a correct manufacturer type in header to mute errors about missing OTA in the repo.


Koenkk:
thanks!

mrmaximas:
@devbis hi! it is possible to flash LYWSD03MMC with pvvx fw to yours? how to do it right? I've tried several ota_index options, but couldn't change the firmware to yours. Thanks!

mrmaximas:
@devbis hi! it is possible to flash LYWSD03MMC with pvvx fw to yours? how to do it right? I've tried several ota_index options, but couldn't change the firmware to yours. Thanks!
Screenshot 2024-11-12 at 20 54 42

devbis:
@mrmaximas You can use a file from this comment https://github.com/devbis/z03mmc/issues/118#issuecomment-1942371299 and use local OTA json file

mrmaximas:
@devbis thanks! it's works! are you planning to implement the measurement interval setting like pvvx?

devbis:
> are you planning to implement the measurement interval setting like pvvx?

I don't have such plans, it is incompatible with zigbee reporting configuration in general

PR #448, opened 2024-02-23
Update index.json for Kobold

WhistleMaster:
Problem: Kobold firmware in the index.json was not correct with a reference to "UPAktorApp" which is not related Kobold.
Solution: Updated index to directly take the firmware from dresden-elektronik.de.


Koenkk:
thanks!

PR #454, opened 2024-03-08
Support OTA updates for multiple Schneider Electric devices.

rmcleay:
These are rebadged TuYa equipment, and FQDNs to the TuYa update servers have been added to the repository rather than the images.

The following devices are now supported with the most recent OTA update available as of 2024-03-08:

- CCT5010-0001 (Micro module dimmer)
- CCT5011-0001 (Micro module switch)
- 2AX (Mech)
- 10AX (Mech)
- Wiser 40/300-Series Module Dimmer (Mech)
- 3025CSGZ (Dual Smart GPO)
- Wiser 40/300-Series Module AC Fan Controller

### DIFF
--- a/index.json
+++ b/index.json
@@ -3319,5 +3319,53 @@
         "sha512": "2a7a17f348f6631e10217d689456ddeceaf768e06267a76a93335a8fb0ee57dfda88e84830536eb478977e423d3add9ff0590839b8736e465e216d0d81f474e9",
         "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/DIY/db15-0203-11003001-z03mmc.zigbee",
         "path": "images/DIY/db15-0203-11003001-z03mmc.zigbee"
+    },
+    {
+        "fileVersion": 34212095,
+        "fileSize": 337195,
+        "manufacturerCode": 4190,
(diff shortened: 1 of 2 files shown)


rmcleay:
Just to clarify (as you can see if you view the the force-push commit above), the two image URLs for the micro module dimmer and the module dimmer have an identical update - just with different URLs for some reason.

This is why I have listed 7 devices in the PR and commit message but only 6 additions are made to index.json.

Koenkk:
Thanks!

rmcleay:
@chrisallen - I can confirm that I've tested it on the dimmer mech (41EPBDWCLMZ/354PBDMBTZ). It does look like there's a bug here somewhere though. Have a look at this screenshot - even after I click on check for updates, it still says that there isn't an update for the lower mech in the screenshot:

Screenshot 2024-03-30 at 09 41 29

Not sure why this is occurring as the most recent firmware is 002.003.007.

rmcleay:
If they're not showing up in the OTA list, then you may not be running the latest version. You can check in  settings -> about. 

What is your Zigbee2MQTT version, and zigbee-herdsman-converters version?

PR #474, opened 2024-04-17
Update NodOn Files

AlexisPolegato:
Update some NodOn Files


Koenkk:
Thanks!

pomeloy:
@AlexisPolegato The inputs/outputs are switched around on my SIN-4-2-20 after installing this firmware, e.g. S1 now switches L2 instead of L1 and vice-versa.

Issue #483, opened 2024-04-23
GL-C-006P(normal) received GL-C-006P(mini) OTA from Z2M


marce14-git:
GL-C-006P (normal version) received the OTA for the MINI Version an is now unusable - all commands are received but light does not turn on etc.

GL-C-006P (normal) has now the GL-C-006P (mini) Firmware = Build Date: 20240227 Version: 20451203.

Manual OTA Update with the right firmware: `"url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/Gledopto/GL-C-006P_10576802_20240412.ota"` just says no firmware update available

see #https://github.com/Koenkk/zigbee2mqtt/issues/22287

[
![image](https://github.com/Koenkk/zigbee-OTA/assets/92689239/da786753-ccd3-424d-97a8-4ed328d2fda2)
](url)
[
![image](https://github.com/Koenkk/zigbee-OTA/assets/92689239/737bd675-36e3-45c0-8f7e-af0d8e3a0195)
](url)

mariustt:
I have exactly the same problem. Additionally, I received the latest OTA version of GL-C-006P (not the mini) from the supplier - https://drive.google.com/file/d/1hA7h_qHQbrTFgNz0aJVS8HNuxzao8MZa

marce14-git:
the root of the problem is that GL-C-006P and GL-C-006P(mini) are not distinguished as seperate entities in Z2M and a new FW for either one of the devices shows up for both.

marce14-git:
> I have exactly the same problem. Additionally, I received the latest OTA version of GL-C-006P (not the mini) from the supplier - [GL-C-006P_V20451203_KEY_20240227.zip](https://github.com/Koenkk/zigbee-OTA/files/15125580/GL-C-006P_V20451203_KEY_20240227.zip)

your firmware did not work because it was the same Version Number than the one the device had - i was able to resolve my problem by emailing Gledopto and getting a new FW see #22287

mariustt:
Thank you for the information, @marce14-git !
I have tried your OTA, the update was successful, but the lights still do not turn on/off. Didn't you do an additional factory reset?

I just want to point out that your OTA is not the latest. I have 9 led drivers. 6 of them I bought at the end of 2022, 3 at the end of 2023. Buy the 2022 version - 10651203, and buy the 2023 version - 20351203.

[...]

PR #484, opened 2024-04-23
Revert "GL-C-006P(Mini)"

Koenkk:
Reverts Koenkk/zigbee-OTA#481 because of https://github.com/Koenkk/zigbee-OTA/issues/483


Koenkk:
CC: @catduckgnaf

catduckgnaf:
> CC: @catduckgnaf 

So what is the best way to fix this if any? Given not as popular do we leave it out?

Maybe we leave in the firmware files if others want to do manually? I was able to apply my firmware. The mini does seem to report as a non mini.

marce14-git:
i was able to fix this with a new Firmware from Gledopto #https://github.com/Koenkk/zigbee2mqtt/issues/22287#issuecomment-2080444066

The Root Cause it that GL-C-006P and GL-C-006P(mini) are not seen as 2 different devices from Z2M OTA so a FW Update for one shows up for both and may brick the other device

PR #504, opened 2024-06-03
Add firmware files version 62 - fixed the bug with freezing values.

BurchuMikhail:
Add firmware files version 62 - fixed the bug with freezing values and removed the disabling of UART communication when the network is unavailable.


Koenkk:
Thanks!

PR #517, opened 2024-06-29
Legrand Firmware Update from Official Legrand Portal (June-2024)

CodeForLove83:
Add Legrand firmware updates for NLF, NLL, NLM, NLP, NLT, NLV, NLC and NLD devices (latest update from official Legrand portal / Last update June 06, 2024)


Koenkk:
Thanks!

CodeForLove83:
@Koenkk I guess there's a problem with the OTA update because of the field `modelId` in the file `index.json`.

In the previous version of the file, Legrand devices didn't have this field and thus in the function `getImageMeta` (file `zigbeeOTA.js`) the if statement is true because of `!i.modelId` and the update can proceed. With the automatic generation of the `index.json` through `node scripts/add.js` the `modelId` is generated for each entry to `"modelId": "done"` and, as the if statement is not `true`, the update is failing.

```
async function getImageMeta(current, device) {
    logger_1.logger.debug(`Getting image metadata for '${device.modelID}'`, NS);
    const images = await getIndex();
    // NOTE: Officially an image can be determined with a combination of manufacturerCode and imageType.
    // However Gledopto pro products use the same imageType (0) for every device while the image is different.
    // For this case additional identification through the modelId is done.
    // In the case of Tuya and Moes, additional identification is carried out through the manufacturerName.
    const image = images.find((i) => i.imageType === current.imageType && i.manufacturerCode === current.manufacturerCode &&
        (!i.minFileVersion || current.fileVersion >= i.minFileVersion) && (!i.maxFileVersion || current.fileVersion <= i.maxFileVersion) &&
        (!i.modelId || i.modelId === device.modelID) && (!i.manufacturerName || i.manufacturerName.includes(device.manufacturerName)));
    if (!image) {
        return null;
    }
    return {
        fileVersion: image.fileVersion,
        fileSize: image.fileSize,
        url: image.url,
        sha512: image.sha512,
        force: image.force,
    };
}
```

I haven't tried yet but the easiest solution would be manually deleting the `modelId` in the file `index.json`.

Before PR I just wanted checking with you.

Issue #527, opened 2024-07-15
Bticino/Legrand NLL devices reporting onboard firmware being newer than latest version


nomc2:
Hi,
My Bticino K4003C which are Legrand NLL devices are reporting onboard firmware being newer than the latest firmware online.
Below an example update request:
```
[2024-07-15 22:24:43] debug: 	z2m:mqtt: Received MQTT message on 'zigbee2mqtt/bridge/request/device/ota_update/check' with data '{"id":"Luce Cameretta","transaction":"seuxf-99"}'
[2024-07-15 22:24:43] debug: 	zhc:ota:common: Checking if an update is available for '0x0004740000b23680' ( Light switch with neutral)
[2024-07-15 22:24:43] debug: 	zhc:ota:common: Using endpoint '1'
[2024-07-15 22:24:43] debug: 	zh:controller:endpoint: CommandResponse 0x0004740000b23680/1 genOta.imageNotify({"payloadType":0,"queryJitter":100}, {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":1,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false,"sendPolicy":"immediate"})
[2024-07-15 22:24:43] debug: 	z2m: Received Zigbee message from 'Luce Cameretta', type 'commandQueryNextImageRequest', cluster 'genOta', data '{"fieldControl":0,"fileVersion":4015615,"imageType":15,"manufacturerCode":4129}' from endpoint 1 with groupID 0
[2024-07-15 22:24:43] debug: 	zhc:ota:common: Got request '{"fieldControl":0,"manufacturerCode":4129,"imageType":15,"fileVersion":4015615}'
[2024-07-15 22:24:43] debug: 	zhc:ota:common: Is new image available for '0x0004740000b23680' ( Light switch with neutral), current '{"fieldControl":0,"manufacturerCode":4129,"imageType":15,"fileVersion":4015615}'
[2024-07-15 22:24:43] debug: 	zhc:ota: Getting image metadata for ' Light switch with neutral'
[2024-07-15 22:24:43] debug: 	zhc:ota: Downloaded main index
[...]

PR #529, opened 2024-07-18
Fix: Legrand June 2024 update

FabianMangold:
* Housekeeping: Remove old binaries
* Fix version reporting for NLL, NLP, NLV

With ref. to https://github.com/Koenkk/zigbee-OTA/issues/527#issuecomment-2231831709, i am happy to patch other images if required. I cannot patch them all for now, since it's been determined that some other product lines work fine (i.e. are not affected by this issue)!
The ones above were tested on my own devices, and can safely be released.


Koenkk:
thanks!

PR #530, opened 2024-07-21
Bosch Plug Compact EU (`BSP-FZ2`): Add firmware 2.10.00 from July 15th

DerDreschner:
On July 15th, Bosch released the firmware 2.10.00 for the `BSP-FZ2` (also known as Plug Compact EU). I was able to sniff and extract the update file as well as update several plugs over Z2M afterwards. Please don't mind the different naming scheme of the ota file. This happened due to using  https://github.com/zigpy/zigpy-cli instead of https://github.com/compujuckel/ZigbeeOtaExtractor for extraction.

Changelog can be found at https://www.bosch-smarthome.com/uk/en/release-notes/bosch-smart-home-system/2024-07-15.html and is copied here for completeness:

* New firmware, which closes a possible security gap, is available for the Dimmer, Relay, and Smart Plug Compact devices.


Koenkk:
Thanks!

PR #531, opened 2024-07-30
Gledopto GL-C-203P

dinamik1488:
Please add new GL-C-203P_V20851203_20240717.ota
As the Chinese said, stateful support for cct has been added


index.json
```diff
@@ -3740,5 +3740,14 @@
         "sha512": "c5145d4c73e7ef61d04558c209e03bf990f50a7589385c1cd662a1a95a2183a31771b9ad38592207339b2d67245f4bf9f01957f1da59817426aac4e007eead1c",
         "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/Legrand/1021-0034-00064203-NLY.zigbee",
         "path": "images/Legrand/1021-0034-00064203-NLY.zigbee"
+    },
+    {
+        "fileVersion": 671166465,
```
Koenkk, line 3745: Please add the model id as also done by other Gledopto OTAs: https://github.com/Koenkk/zigbee-OTA/blob/32b365409415cad741d381781176b7e49d4c49e7/index.json#L938

dinamik1488: ?

dinamik1488: The GL-C-203p device is 2 devices in 1, they are defined as 006p and 009p. Switch from 1 to another. The Chinese said that this firmware is for 009P.

Koenkk: I'm a bit worried that this OTA will also be applied to `GL-C-009P` and not just the `GL-C-203P` which may cause devices to get bricked (example: https://github.com/Koenkk/zigbee2mqtt/issues/23128)

dinamik1488: How can I check?

Koenkk: I think the only way to check is with an actual `GL-C-009P`

PR #535, opened 2024-08-05
Update SIN-RS Firmware

AlexisPolegato:
Rollback to previous version on SIN-RS Firmware 2 as it seems to generate some issues in the OTA


Koenkk:
Thanks!

PR #553, opened 2024-09-05
Add firmware files version 63 for wb msw zigbee v4   fix low transmitter power during operation

BurchuMikhail:


Koenkk:
Thanks!

PR #554, opened 2024-09-12
Add new Innr OTAs

Koenkk:
Release notes:

*   Fixed scene handling issue where in some cases the transition time was overwritten with a large value.
    
*   Fixed handling of transition time 0xffff in Level Control commands.
    
*   Improved attribute updates in Level Control and Color Control clusters for simultaneous transitions with time > 0.
    
*   Fixed attribute reporting after power cycle.
    
*   Corrected response codes of several command results in Scenes, Level Control and Color Control clusters.
    
*   Corrected stack revision in Node Descriptor.
    
*   Corrected Basic cluster "reset to factory new" command.
    
*   Increased broadcast transaction table size from 20 to 25.
    
*   Send permit join after joining (Zigbee BDB compliance).
    
*   Added Device Announce when starting up in networked mode.
    
*   Improved blinking indications for factory-new and touchlink scenarios.
    
*   New chipset SDK version:
    
    *   Fix buffer free issue in case incoming counter of APS encrypted data doesn't match.
        
    *   RREQ only retransmitted if the radius is greater than 0.
        
    *   Various other small fixes and improvements.


evelant:
Hooray my zigbee network is finally stable and responsive! The OTA fixed all issues I was having with Innr AE270T bulbs. Thank you!

Issue #420, opened 2023-12-29
`https://otau.meethue.com` started using a self-signed certificate


kaechele:
Signify seems to have started using self-signed certificates for `https://otau.meethue.com`.
Unsure if this is a bug on their end or a new configuration they have adopted going forward.

Either way, currently OTA updates using images hosted on that URL will fail:
```
debug 2023-12-29 05:55:33: Received MQTT message on 'zigbee2mqtt/bridge/request/device/ota_update/update' with data '{"id":"Bedroom Lamp","transaction":"r82z7-1"}'
info  2023-12-29 05:55:33: Updating 'Bedroom Lamp' to latest firmware
debug 2023-12-29 05:55:33: Received Zigbee message from 'Bedroom Lamp', type 'readResponse', cluster 'genBasic', data '{"dateCode":"20230606","swBuildId":"1.108.5"}' from endpoint 11 with groupID 0
info  2023-12-29 05:55:33: MQTT publish: topic 'zigbee2mqtt/Bedroom Lamp', payload '{"brightness":254,"color":{"h":32,"hue":32,"s":81,"saturation":81,"x":0.4574,"y":0.41},"color_mode":"color_temp","color_temp":366,"color_temp_startup":65535,"last_seen":"2023-12-29T10:55:33.511Z","linkquality":144,"power_on_behavior":"previous","state":"OFF","update":{"installed_version":16786946,"latest_version":16786948,"state":"available"},"update_available":null}'
debug 2023-12-29 05:55:33: Updating to latest '0x0017880103adfb62' (LCT016)
debug 2023-12-29 05:55:33: Using endpoint '11'
debug 2023-12-29 05:55:33: Received Zigbee message from 'Bedroom Lamp', type 'commandQueryNextImageRequest', cluster 'genOta', data '{"fieldControl":0,"fileVersion":16786946,"imageType":268,"manufacturerCode":4107}' from endpoint 11 with groupID 0
[...]

PR #555, opened 2024-09-16
Update Hue CA certificates

kaechele:
Signify is playing around with self-signed CA certs again. Update the CA cert bundle to fix OTA updates, as reported in #420.


Koenkk:
Thanks!

PR #560, opened 2024-09-20
Add new firmware for Aqara SP-EUC01 0.0.0_0045

inode64:
Changelog from app
----
Fix known issues


Koenkk:
Thanks!

TheJulianJES:
We should confirm if v45 fixes https://github.com/Koenkk/zigbee2mqtt/issues/13903. Otherwise, that bug is reintroduced with this PR.
Similar discussion in previous PR that wanted to add v43: https://github.com/Koenkk/zigbee-OTA/pull/278

Marlor:
This firmware should not be used
"random" triggering is back, it even triggers if a SONOFF SNZB-02P joins the zigbee network in the vicinity.

PR #564, opened 2024-10-04
update CSM-300ZB ota file to version 15

shinasys:
. v15 release note: During maintaining a full charge over 10 minutes, the TOF sensor is always activated, allowing operation without a PIR motion sensor.


Koenkk:
Thanks!

PR #565, opened 2024-10-07
New FW for Namrom 1402769 dimmer

TurboJens:
Added FW file for Namron dimmer. Should been under Namron, not EcoDim - or both?

Changes related to Jennic/Osram was initiated by the script.


index.json
```diff
@@ -80,12 +80,12 @@
...
         "minFileVersion": 1107326256,
-        "fileVersion": 1124103171,
-        "fileSize": 256696,
+        "fileVersion": 1107326256,
+        "fileSize": 256632,
         "manufacturerCode": 4107,
         "imageType": 260,
-        "sha512": "eb9e81b28ea8128831c0f656e65be2821b6d06207bd44adbc31b050ed41e3656edc10e1c0a27cf73a2817c257208f9002c3e219913903ecdaacf7857f799b001",
-        "url": "https://otau.meethue.com/storage/ZGB_100B_0104/33cbfd3d-3b58-43e2-a6a0-1e6fe50f2bee/ConnectedLamp-Atmel-Target_0012.sbl-ota"
+        "sha512": "d2bf330b9a23114efb6a613ccefce691e4a67a98175033e65d3eaf6841312b5a542bd538ae19c06c3804aa06224d35acc184f4b37a6198b457df2a173a490f21",
```
Koenkk, line 87: Why did this entity change?

TurboJens: I really don't know. I just added the file in a fresh fork (uploaded it, since url gave error 403 Forbidden), run node scripts/add.js PATH_TO_OTA_FILE and then run node scripts/updateall.js and these changes showed up. I did not touch anything else.

I can start all over again and see if I get another result?

Koenkk: `updateall.js` shouldn't be needed, only `add.js

TurboJens: Damn, I did not read the instructions properly..

I will redo it from scratch, but before we begin I thougt I should test if it works since add.js generates a code with wrong manufacturer ID. So I added firmware in local OTA index and tried with different settings without anything seem to work. I tested added another firmware and by modifying the "file version" it responded with a available upgrade so I know the local index is working.

But, a question, when a device does not show up in the OTA tab of z2m, is that because it does not support OTA at all? Comparing to another supported device from the same vendor I could not se any difference in clusters.

Thanks.

Koenkk: It's because the definition is missing `ota: ota.zigbeeOTA` ([example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/b53aef5c1d93f33f68835ea96d2902c5a93e8979/src/devices/namron.ts#L334))

TurboJens: Yeah, I found it.. so I played around a bit and made my own external definition for this model. I have 6 ones that is working perfect, and 5 that is not working as intended.

When I enable OTA i get updates for the ones that is working (without adding the firmware file locally) and no update for the test unit that I want to upgrade..

[14-10-242024 01:16:12] info: 	z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Testdimmer', payload '{"brightness":254,"linkquality":58,"power_on_behavior":null,"state":"OFF","update":{"installed_version":1,"latest_version":1,"state":"idle"},"update_available":null}'
Old/new version is 1

[14-10-242024 01:16:11] info: 	z2m:mqtt: MQTT publish: topic 'zigbee2mqtt/Soverom master', payload '{"brightness":35,"level_config":{"on_level":53},"linkquality":134,"power_on_behavior":null,"state":"ON","update":{"installed_version":2,"latest_version":4,"state":"available"},"update_available":null}'
The working one is on version 2 and can update to 4.. I have no idea what file that's made available as there should not be a firmware for this in zigbee-OTA and the one I have downloaded from manufacturer is version 2. Any idea?

They seem to be identical except from the date code, all with the latest date code is workin, all with the 2022 datecode is failing (does not report back status when using the dimmer wheel, at least for ZHA, since they did not work I replaced them so the test unit does not have a load connected, was trying to make them work.

image

TurboJens: `    {
        "fileVersion": 4,
        "fileSize": 286646,
        "manufacturerCode": 4714,
        "imageType": 1000,
        "sha512": "9b7835e32ef5af35385710332838099c5dce99a398ae04a1a769d67935fb0752ca4da58892f13bc842cbf8f9ae20516142fe9b66eff3e8ebef6577a71a1d9b0b",
        "url": "https://github.com/Koenkk/zigbee-OTA/raw/master/images/EcoDim/ZB21S3_HZC_Dimmer1_EcoDim-Zigbee%203.0_1.01_20230908.ota",
        "path": "images/EcoDim/ZB21S3_HZC_Dimmer1_EcoDim-Zigbee 3.0_1.01_20230908.ota"
    },`

Could maybe be this one.. 1 og 3 with fileVersion 4 in the index. It's placed under EcoDim just as the file I uploaded. I would love to try updating one but I'm afraid I'll destroy one that's working so best to leave it...

IDmedia: I brought 11 of these dimmers and they have been a mess to work with in z2m. I've also noticed a new firmware is available, but I haven't been able to update using a local json file (and I'm not sure if i should).
I'm currently on the same 3.09 firmware, but get this error all the time. Have you encountered the same issue @TurboJens ?

image

PR #566, opened 2024-10-07
Legrand 09122024

CodeForLove83:
Updated all the Legrand firmware files from Legrand Development Portal.
Latest update: 12 September 2024


Koenkk:
Thanks!

PR #569, opened 2024-10-15
Add new firmware v1.1.0 for TS011F/_TZ3000_fukaa7nc

ppluciennik:
Downgrading from v3.0.0 to v1.1.0 seems to fix random turnoff when plug detects under voltage.

Plug reports under voltage, but doesn't turnoff.


Koenkk:
Thanks!

PR #592, opened 2024-11-01
Fix extra metas retrieval for push context.

Nerivec:
Since moved from pull_request to push context for "update after merged OTA", extra metas were no longer passed properly in update manifests workflow.
Added some tests for extra coverage.

Note: This requires the message during merge to be properly set with the PR number, as is the default message, with format `(#...)`
merge-msg
_There might be a better way, but the PR isn't passed to the push context..._


Koenkk:
Thanks!

PR #599, opened 2024-11-12
Add latest firmware images for NodOn devices

AlexisPolegato:
Add Firmware V3.5.0 for:
- SIN-4-1-20
- SIN-4-1-21
- SIN-4-2-20
- SIN-4-FP-21
- SIN-4-RS-20


Nerivec:
Please move the images to the proper subfolder under `images`.
https://github.com/Koenkk/zigbee-OTA?tab=readme-ov-file#adding-new-and-updating-existing-ota-files

Koenkk:
Thanks!

rclsilver:
I also have a SIN-4-FP-21_EQU module, and in my case, it no longer reports power consumption (neither instantaneous nor cumulative). I reached out to NodOn support last week, and they mentioned that this issue might be resolved with the latest firmware update. They told me they would try to push the updates soon.

Hoping this can help move things forward.

Thanks!

molusk:
> @AlexisPolegato , seems you're working at Nodon. Could you please push a firmware update for SIN-4-FP-21_EQU ? Those modules have never been updated and lacks behind SIN-4-FP-21 updates. They still have many bugs that are already fixed on SIN-4-FP-21. Hardware is the same.
> 
> Ota has been enabled for those modules on latest z2m release, we just need some fw files now.
> 
> I was in contact with Enki support team, they says they can't OTA on their zigbee box anyway, and the product is not from them but NodOn, they're not doing support on it.
> 
> Thx.

I was so happy to find this module in my local store I just checked the price was the same as the nodon one and bought it...
Maybe should I rather order a real nodon module and send back the _EQU one to avoid issues ? (I am still in the free return period)

jbjardine:
Following the upgrade of Zigbee2MQTT to version 2.x, OTA updates for the node-on SIN-4-FP-21 firmware are consistently failing with an “Image CRC-32 is invalid” error. This issue appears to be isolated to this firmware image, as other firmware updates proceed normally. It remains unclear whether the problem stems from changes in the CRC32 verification logic in Zigbee2MQTT 2.x or if the firmware image itself is corrupted. Further investigation is needed.
Thanks!

Issue on Zigbee2MQTT :
[https://github.com/Koenkk/zigbee2mqtt/issues/25699](url)

PR #607, opened 2024-11-21
Bosch BTH-RA (Radiator thermostat II): Add firmware 3.06.08 from November 20th, 2024

DerDreschner:
Yesterday, Bosch released a new firmware for their radiator thermostat II. It has the version number 3.06.08 and include several changes to their heating algorithm [according to their changelog](https://www.bosch-smarthome.com/uk/en/release-notes/bosch-smart-home-system/2024-11-20.html).

> Radiator thermostat II
> The heating control algorithm has been revised at some points to address specific states in the wide range of heating systems more effectively. The set desired temperature is now reached and maintained more quickly.

As they don't release an official download, I had to sniff the firmware once again. To test the package, I create this PR and try the update file with my second thermostat. That's the reason it's still a draft.

Enjoy!


DerDreschner:
@mmattel : Fingers crossed! 🤞 I use "Better Thermostat", therefore, I have no experience with the native open-window-implementation. But would be interested to know if it's fixed for you or not.

mmattel:
Sure!
In a nutshell, I have door sensors that trigger on open/close. In HA, I use an automation that gets triggered by the sensors and progresses if the states do. ot change for a defined time. Then I check if heating is not off and trigger window open/close on the bth accordingly. This works process wise totally fine, but the bth somtimes sets back the temp not back to the former but much lower value...

mmattel:
@DerDreschner The update went frictionless, many thanks ! 👍

datenhamster:
Do you have an archive of the old files? The new one (374) triggers an endless calibration.

DerDreschner:
@datenhamster : You can find the old files in the Git commit history. But I'm pretty sure you would have to patch the old firmware to install it on the thermostat as a downgrade may not be possible otherwise.

Nerivec:
When the previous update is in the archive (index1.json), you can use the new MQTT endpoints in Z2M 2.0.0 to trigger a downgrade:
https://www.zigbee2mqtt.io/guide/usage/ota_updates.html#downgrade

Note: due to the nature of that feature, we have very little feedback on it, but as DerDreschner mentioned, it is up to the device to accept or refuse a downgrade, and if the latter, there is nothing Z2M can do about it.

datenhamster:
Thank you. The downgrade with Z2M ist not working. It shows an available "upgrade" after the downgrade-check. But with the same version.

{"installed_version":923014420,"latest_version":923014420,"state":"available"}

With Bosch making the device useless for me i switched on to Sonoff. It works, so I will get rid of the other Bosch devices too.

PR #608, opened 2024-11-23
Add Danalock Zigbee Firmware v.17.7

Schluggi:


Nerivec:
@Schluggi can you confirm that the two firmware mentioned in https://github.com/Koenkk/zigbee-OTA/issues/639 are for the same device?
Can do a manual cleanup as needed (missing `modelId` in firmware added by this PR, so it didn't archive the other one).

Schluggi:
Hi @Nerivec,

There is only one Zigbee based Danalock (the V3). So yes, I can confirm.

The modelID is missing because the new github action based way does not allow to? At least I didn't know how. I thought it was not necessary anymore. Sorry for creating problems. 

Additional update hint: In my case the update failed every time. I had to press the small pairing button at the device before press update. In case you mention a similar problem.

Nerivec:
👍 

For future reference (about adding metadata): https://github.com/Koenkk/zigbee-OTA?tab=readme-ov-file#declaring-extra-metadata-for-automatic-inclusion-in-the-manifest

PR #614, opened 2024-11-28
Revert "Update thirdreality TH lite sensor ota file to v44"

weihuan1111:
Reverts Koenkk/zigbee-OTA#613


Koenkk:
The currently workflow doesn't allow reverting easily so I reverted it in https://github.com/Koenkk/zigbee-OTA/commit/31f4dcf5c92406c68b43f3116b9d27dda95a1655, thanks!

PR #626, opened 2024-12-13
Added the v1.0.4 firmware of ZBMINIR2, adding the function of network…

liubotongs:
… indicator control setting.


liubotongs:
图片
This does not need to update the index.json file.

Nerivec:
You should disable the workflows in your fork to avoid errors on that side. They are meant to be run here.
Commits from fork workflows cannot be included in PRs, it will mess with the workflows here.

Koenkk:
Thanks!

PR #640, opened 2025-01-06
fix MHO-C401 refresh EPD

migo1271:
This is the latest OTA file for device MHO-C401N (0.1.2.3).

```json
{
    "modelId": "MHO-C401N"
}
```


Koenkk:
Thanks!

PR #653, opened 2025-01-14
add firmware version 65 for wb-msw-zigbee-v4

BurchuMikhail:
- firmware update to 4.31.11
- add the ability to change baud rate
- add a flag for the presence of a modbus connection


Koenkk:
Thanks!

PR #655, opened 2025-01-21
Add firmware for Namron 1402769

IDmedia:
Not sure why this was never added since there was a pull earlier (https://github.com/Koenkk/zigbee-OTA/pull/565), but I still hope this new firmware can be added.


Koenkk:
Thanks!

TurboJens:
It's working, but I have a older batch of these dimmers and it will not update them. I have earlier tried  #565 to investigate the differences between these two without luck. Unfortunatly the ones that is updating now is working fine, the oldest have some issues with reporting back the dimmer level when using the dimmer wheel manually so they will certainly benefit from a  update.
image

IDmedia:
I updates all my dimmers and they seem to work fine in Home Assistant, but every time I try to pair a new device I get a ton of these messages filling up the screen

error

I don't quite understand the versioning either here, the dimmer's build date and version (3.09) didn't change, but I can see that this did from 2 to 4. Must be rocket science to do versioning right in Zigbee :p

image

TurboJens:
So, what you say is that the dimmers will work fine with version 4, but if you for any reason need to pair it again it won't work anymore?

IDmedia:
> So, what you say is that the dimmers will work fine with version 4, but if you for any reason need to pair it again it won't work anymore?

I didn't have any issues controlling them before or after the update. The only problem is that they throw that error message every now and then, especially when pairing a new device. I'm also struggling with half of my 60+ devices going offline and I suspect the Namron's are to blame, but I have no evidence to back up that claim with yet. Didn't have any issues before installing 11 of these.

TurboJens:
> > So, what you say is that the dimmers will work fine with version 4, but if you for any reason need to pair it again it won't work anymore?
> 
> I didn't have any issues controlling them before or after the update. The only problem is that they throw that error message every now and then, especially when pairing a new device. I'm also struggling with half of my 60+ devices going offline from time to time. I suspect the Namron's are to blame, but I have no evidence to back up that claim with yet. Didn't have any issues before installing 11 of them.

I have 6 of them, plus a test dimmer that is not in use. I have no issues or error messages in z2m.

I do notice that the "Firmware build date" actually did change, from 20230401 to 20230908. It's still old so I did not notice at first.

IDmedia:
I just noticed that some may have changed. I have now 6 with 20230908 and 5 with 20230401. I think these build numbers are cached somehow and not necessarily updated. I also own a Danalock v3 and even though a friend of mine and my own are on the same FW, the build date differs.

TurboJens:
> I just noticed that some may have changed. I have now 6 with 20230908 and 5 with 20230401. I think these build numbers are cached somehow and not necessarily updated. I also own a Danalock v3 and even though a friend of mine and my own are on the same FW, the build date differs.

All of mine have changed, I have screenshot of the OTA page before and during update. Now I have updated all 6 that will allow update and they all changed from 20230401 to 20230908.

PR #659, opened 2025-01-24
feat: added v3.12 firmware for namron 4512750 and 4512751

SanderBlom:
This PR adds the latest firmware for the Namron dimmer 4512750 (white edition) and 4512751 (black edition).
This firmware is only supported for the devices sold from late summer 2024 and that is running the current 3.10 firmware.

Firmware was sourced here: https://www.hjemmeautomasjon.no/forums/topic/12466-namron-dimmer-45127514512750-problemer/#comment-117469. The firmware was posted to the forum by a verified Namron employee as the thread discusses the current [issue](https://github.com/Koenkk/zigbee2mqtt/issues/24422) with dimmers running the v3.10 firmware not responding to commands.


Koenkk:
thanks!

PR #665, opened 2025-02-01
Update LICENSE

Koenkk:
Based on the discussion from https://github.com/Koenkk/zigbee-OTA/pull/663

CC: @Nerivec @mundschenk-at @TheJulianJES


TheJulianJES:
I think this is a good step. I guess this is somewhat similar to the HA brands repo (with no license) and a copyright disclaimer then: https://github.com/home-assistant/brands#trademark-legal-notices

PR #670, opened 2025-02-02
Update LiXee OTA Files

fairecasoimeme:


Nerivec:
LiXee is part of the weekly auto download. No need for manual PR, it's all updated automatically.

PR #679, opened 2025-02-06
Fix url encoding

Koenkk:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/26210

Note that I ran a reprocess before (https://github.com/Koenkk/zigbee-OTA/pull/678)


Koenkk:
I'm not sure, you basically have to disallow any character that has to be URL encoded. This seems like a robust solution.

Issue #697, opened 2025-02-18
Run autodl more regularly


rohankapoorcom:
If a firmware file is uploaded by the manufacturer on a Monday, it seems weird to wait nearly a full week for the Sunday run to grab it. Can we run the process 1x / day instead?

Koenkk:
Im fine im with this, @Nerivec any objections?

Nerivec:
I'd rather keep the process weekly. It keeps the requests amount to various external servers lower, and in 99% of cases, it won't make any difference to run it more often anyway.
If an urgent update becomes available, we can trigger it manually, like we've already done in the past.

rohankapoorcom:
> I'd rather keep the process weekly. It keeps the requests amount to various external servers lower

Is this really an issue though? Before the updates were centralized here, each users Z2M reached out to the companies servers every time a check OTA update was hit. This is still a single centralized set of requests going to the external servers.

PR #708, opened 2025-02-28
Run auto download daily for specific manufacturers

rohankapoorcom:
As discussed with @Nerivec in the context of #697, a lot of manufacturers have extensive rate limiting on their API endpoints. For manufacturers that specifically request their OTAs being processed more frequently, we can run a custom job for this one.

Added a new action that will run daily and has an environment variable to populate the list of manufacturers. It intentionally does not trigger a release (keeping that for the weekly job)

Fixes: #697

### DIFF
--- a/.github/workflows/run_autodl_daily.yml
+++ b/.github/workflows/run_autodl_daily.yml
@@ -1,0 +1,51 @@
+name: Run auto download
+on:
+    schedule:
+        - cron: '0 0 * * *'
+
+env:
+    # comma separate values, no spaces
+    MANUFACTURERS_TO_PROCESS: "inovelli"
+
+permissions:
+    contents: write
+
+jobs:
+    run-autodl:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+
+            - uses: pnpm/action-setup@v4
+              with:
+                  version: 9
+
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: 20
+                  registry-url: https://registry.npmjs.org/
+                  cache: pnpm
+
+            - name: Install dependencies
+              run: pnpm i --frozen-lockfile
+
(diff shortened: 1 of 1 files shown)


.github/workflows/run_autodl_daily.yml
```diff
@@ -0,0 +1,51 @@
+name: Run auto download
+on:
+    schedule:
+        - cron: '0 0 * * *'
```
Nerivec, line 4: Can you offset the time to mid-day or something, so it doesn't trigger same time as the regular weekly one?

.github/workflows/run_autodl_daily.yml
```diff
@@ -0,0 +1,51 @@
+name: Run auto download
```
Nerivec, line 1: `Run daily auto download`

.github/workflows/run_autodl_daily.yml
```diff
@@ -0,0 +1,51 @@
...
+                      const {runAutodl} = await import("${{ github.workspace }}/dist/ghw_run_autodl.js")
+
+                      await runAutodl(github, core, context, "$MANUFACTURERS_TO_PROCESS")
+
+            - name: Commit changes
+              run: |
+                  git config --global user.name 'github-actions[bot]'
+                  git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+                  git add .
+                  git commit -m "Autodl update" || echo 'Nothing to commit'
```
Nerivec, line 50: `"Daily autodl update"`

so we can easily differentiate the two.

.github/workflows/run_autodl_daily.yml
rohankapoorcom: Moved to 12:00

rohankapoorcom:
Changes implemented, thanks!

PR #717, opened 2025-03-07
Ubisys S1 missing firmware

FlorianBruckner:
Ubisys S1 has identical versions for different hardware revisions. Currently, only the old hardware revisions are in the catalog. With this fix, hardware version constraints are taken into account when checking for existing firmware.


FlorianBruckner:
Tests are adjusted to take the new behaviour for images with hardware constraints in source metadata.

Koenkk:
Thanks!

FlorianBruckner:
HI @Koenkk,

just so you're aware: This causes an issue in zigpy and Homeassistant because it cannot handle multiple images with the same version where the hardware constraints are different. Having Homeassistant and ubisys devices lead me to create this PR in the first place. They are actually checking hardware constraints before starting an OTA and fail, but it turns out other parts of that OTA implementation have an issue with this and refuse to offer any OTA for these devices now.

Nerivec:
Problem is, you have:
`device.hardwareVersion`
`extraMetas.hardwareVersionMin` / `extraMetas.hardwareVersionMax`
`header.minimumHardwareVersion` / `header.maximumHardwareVersion`

All of which should point to the same underlaying value representing `device.hardwareVersion`. But are the values present in the Ubisys manifest actually the values used in the OTA file, and in the actual hardware?

PR #719, opened 2025-03-11
frient Smoke/Heat alarm firmware update

fsedarkalex:
Just got this from the manufactuer. It should **fix** a firmware corruption issue present in the 4.0.8 update causing failing devices


fsedarkalex:
Probably worth to note: the "smoke alarm" and "heat alarm" share the same firmware according to the manufacturer

PR #727, opened 2025-03-23
OTA image for Tuya thermostats with custom firmware

slacky1965:
```json
{
    "manufacturerName": ["Slacky-DIY"]
}
```


slacky1965:
> What is the manufacturer name of these devices?

These are thermostats based on the Tuya platform, but reprogrammed using custom firmware. I use Slacky-DIY.

Koenkk:
Could you provide the data/database.db entry when paired with z2m?

slacky1965:
> Could you provide the data/database.db entry when paired with z2m?

Ok. Can't attach *.db file. Put in my server.

[database.db](https://www.slacky.ru/files/database.db)

Koenkk:
It has `"manufacturerName":"Slacky-DIY"` right?

slacky1965:
> It has `"manufacturerName":"Slacky-DIY"` right?

Yes.

Koenkk:
Thanks! Make sure the definition in zigbee-herdsman-converters also had `ota: true`

PR #739, opened 2025-04-07
feat: keep records of identifiable stacks for firmware files

Nerivec:
Purely for dev usage.
Knowing the stacks running on certain devices when debugging can be pretty useful sometimes. This is limited by the availability of OTA firmware files of course, and the "possibility" to identify stacks from them, but it still provides a decent database.
_Kept as a separate JSON to not crowd the manifests._
I've looked up a few of the magic numbers that can be used to identify stacks, but still some left to do (closed source makes that harder...).
Set to run with autodl workflow.

Also cleaned up null chars from manifests (and generating code).


Koenkk:
Nice, thanks!

PR #740, opened 2025-04-07
Bosch BTH-RA (Radiator Thermostat II): Add firmware 3.07.06 from April 2nd, 2025

DerDreschner:
Last week, Bosch released a new firmware for their Radiator Thermostat II. It's version number 3.07.06 and fixes a possible endless loop during auto-calibration (see [changelog](https://www.bosch-smarthome.com/uk/en/release-notes/bosch-smart-home-system/2025-04-02.html)).

As they don't release an official download, I had to sniff the firmware once again. Unfortunately, I were unable to test the package myself, as I had to update both of my thermostats due to missing frames according to zigpy-cli when sniffing the update on my first thermostat. But that way, I'm even more sure the file is fine.

```json
{
    "modelId": "RBSH-TRV0-ZB-EU",
    "releaseNotes": "1. A fault has been resolved that caused the radiator thermostat to sometimes enter an endless loop during auto-calibration"
}
```


Koenkk:
Thanks!

PR #745, opened 2025-04-11
Remove ManufacturerName for Namron

joimaggz:
For some reason when ManufacturerName is present for Namron devices, Z2M does not find new firmware, even though the value matches the name in the device information in Z2M. When it is removed, it finds the firmware. This is tested on my fork.

### DIFF
--- a/index.json
+++ b/index.json
@@ -7600,9 +7600,6 @@
     "manufacturerCode": 4644,
     "sha512": "ff1f08b6e1653484d759d9538593b53b62597c44f4b3038ce581190be8cf1dd858f94ab0a7a462d6c3d462f06e5039b6726671baa318a4277a4a18cb8a1e6234",
     "otaHeaderString": "Encrypted GBL Z3DimSoc_sdk690_ef",
-    "manufacturerName": [
-      "Namron"
-    ],
     "modelId": "4512760"
   },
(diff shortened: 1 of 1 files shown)


Nerivec:
If there's a bug somewhere in the logic, we should fix that instead.
Have you been able to locate where the matching fails exactly, or why?
Can you post the `database.db` entry for the device and some `debug` logs of the failed vs success matching?

joimagg:
Hey, thanks for the advice. I found the manufacturername in database.db and it was 'NAMRON AS'.
I tried that name out in my fork and it works now. The letters are case sensitive so has to be all caps to work.
I will revise the pull request.

From database.db - until "manufacturerName"
{"id":2,"type":"Router","ieeeAddr":"0x943469fffeee31a6","nwkAddr":29235,"manufId":4644,"manufName":"NAMRON AS","powerSource":"Mains (single phase)","modelId":"4512760","epList":[1,242],"endpoints":{"1":{"profId":260,"epId":1,"devId":257,"inClusterList":[0,3,4,5,6,8,1794,2820,2821,4096],"outClusterList":[25],"clusters":{"genBasic":{"attributes":{"modelId":"4512760","manufacturerName":"NAMRON AS"...

PR #746, opened 2025-04-12
Change manufacturerName from "Namron" to "NAMRON AS"

joimaggz:
ManufacturerName was added with wrong name. Correct name is tested in fork and works.

From database.db in M2Q
{"id":2,"type":"Router","ieeeAddr":"0x94346...","nwkAddr":39791,"manufId":4644,"manufName":"NAMRON AS"


Koenkk:
Thanks!

PR #460, opened 2024-03-20
Legrand 20 03 24

SilentT-FR:

PR #751, opened 2025-04-17
Update Legrand images (2025-04-03)

michal-kocarek:


michal-kocarek:
This process is new to me, so I am afraid I've made some terrible mistakes despite going through the docs.

I've downloaded all available images from [Legrand firmware page](https://developer.legrand.com/production-firmware-download/).

However, some of the images, like `NLF` and `NLL` are present more times (see the screenshot below). 

image

Should I somehow specify minimal and maximal available version for them, include only latest one or what should be the best way forward?

michal-kocarek:
I was looking more at the naming scheme of the firmware. Here's what I found:

```
1021-000f-004145ff-NLL.zigbee
     ^^^^              = imageType (15)
          ^^^^^^^^     = (0x004145ff) firmware version reported by HomeAssistant
          ^^^^         = (0x0041 = 65) = version on their official website
              ^^^^     = magic number - seems to be either 0x45ff or 0x4203
```

Regarding the magic number, I see these things:

In this repository, we have been switching between `45ff` and `4203` historically for few recent commits. At some point there were both of them for few alternatives.

On the Legrand download page, it looks like this (for firmwares I've downloaded and supplied in this PR):

| Products                                                                                  | Cloud Type | Version | Magic number |
| ----------------------------------------------------------------------------------------- | ---------- | ------- | ------------ |
| Dimmer switch w/o neutral                                                                 | NLF        | 77      | 45ff         |
| Dimmer switch w/o neutral - To be used if current NLF firmware version is lower than 28   | NLF        | 57      | 4203         |
| Light switch with neutral                                                                 | NLL        | 65      | 45ff         |
| Light switch with neutral - To be used if current NLL firmware version is lower than 7    | NLL        | 34      | 4203         |
| Micromodule                                                                               | NLM        | 67      | 45ff         |
| Connected outlet                                                                          | NLP        | 103     | 45ff         |
| See NLT devices list                                                                      | NLT        | 81      | 45ff         |
| Shutter switch with neutral                                                               | NLV        | 62      | 45ff         |
| Shutter switch with neutral - To be used if current NLV firmware version is lower than 21 | NLV        | 47      | 4203         |
| Cable outlet                                                                              | NLC        | 38      | 45ff         |
| Double gangs remote switch                                                                | NLD        | 54      | 45ff         |
| Remote motion sensor                                                                      | NLTS       | 33      | 45ff         |
| Dimmer switch with neutral                                                                | NLFN       | 38      | 45ff         |
| Dimmer switch w/o neutral evo                                                             | NLFE       | 34      | 45ff         |
| Rotary dimmer                                                                             | NLFO       | 11      | 45ff         |
| Rotary color                                                                              | NLOC       | 7       | 45ff         |
| 1/2/3 gangs switch with neutral                                                           | NLIS       | 27      | 45ff         |
| Triple Wireless Switch ON/OFF                                                             | NLW        | 24      | 45ff         |
| 1/2 Gangs Shutter Switch                                                                  | NLIV       | 30      | 45ff         |
| 1 Gang Switch without neutral                                                             | NLIB       | 3       | 45ff         |
| Fan Controller                                                                            | NLH        | 18      | 45ff         |
| Garage Door Module                                                                        | NLJ        | 11      | 45ff         |
| Three-Phase Energy Meter                                                                  | NLY        | 7       | 45ff         |
| Switch in wall – US                                                                       | NLUI       | 45      | 4203         |
| Dimmer Switch in wall – US                                                                | NLUF       | 45      | 4203         |
| Plug ON/OFF – US                                                                          | NLUP       | 45      | 4203         |
| Plug Dimmer – US                                                                          | NLUO       | 45      | 4203         |

---

My PR is now replacing majority of `4203` firmwares with with `45ff`. That is weird. I think we should investigate more why there is so much changes and why this part is different.

Nerivec:
Per ZigBee spec:
> The OTA Upgrade image file name SHOULD contain the following information at the beginning of the name with each field separated by a dash (“-“): manufacturer code, image type and file version. The value of each field stated SHOULD be in hexadecimal number and in capital letter. Each manufacturer MAY append more information to the name as seen fit to make the name more specific. The OTA Upgrade file extension SHOULD be “.zigbee”. An example of OTA Upgrade image file name and extension is “1001-00AB-10053519-upgradeMe.zigbee”.

They seem to follow this, so, looks like you could determine the proper values for `maxFileVersion` with some work 😁
Make sure to `- 1` to have values that pass `<=` (`lower than 28` => `27`).

Note that using the `maxFileVersion` will alter the processing of the PR since it will not replace files that match but with a different maxFileVersion (or no maxFileVersion), so the end result of the PR will only be appropriate once all files are properly tagged with metadata.

SilentT-FR:
I dont know when and why and who but i have already make that

https://github.com/SilentT-FR/zigbee-OTA/blob/adf9354cdbc7a704e66d9ac0fef8b31612c5ba56/index.json#L2723-L2730

https://github.com/SilentT-FR/zigbee-OTA/blob/adf9354cdbc7a704e66d9ac0fef8b31612c5ba56/index.json#L2741-L2750

https://github.com/Koenkk/zigbee-OTA/pull/460/files

SilentT-FR:
> My PR is now replacing majority of `4203` firmwares with with `45ff`. That is weird. I think we should investigate more why there is so much changes and why this part is different.

Legrand never respond to that, and yes when the koenk script read the file, it found a version but when installed another version appear

SilentT-FR:
and between official legrand gateway and the devlopper website you dont have same firmware https://developer.legrand.com/forums/topic/nlp-firmware-not-same-version-from-official-gateway/

https://github.com/Koenkk/zigbee-herdsman-converters/pull/6193#issuecomment-1733258583

alexhk:
Ok, I will give this a shot.

While the developer download page seems incomplete and outdated, there are in fact **only 3 device types** that seem to require "in-between" firmware versions. It's luckily not that many.

https://developer.legrand.com/production-firmware-download/

| Products                    | Type  | Latest Version | If Current Version Below | Use Version Instead |
| --------------------------- | :--------: | :------------: | :----------------------: | :-----------------: |
| Dimmer switch w/o neutral   | NLF        | 78             | 28                       | 57                  |
| Light switch with neutral   | NLL        | 67             | 7                        | 34                  |
| Shutter switch with neutral | NLV        | 63             | 21                       | 47                  |


So lets stick to the NLL for this example. I know that this could  be a single PR for both files, but I'll keep it separate for this discussion. I am looking at [this](https://github.com/Koenkk/zigbee-OTA?tab=readme-ov-file#declaring-extra-metadata-for-automatic-inclusion-in-the-manifest) info.

PR #1171 for v71 could get this in the body:

```
[
    {
        "fileName": "1021-000f-004745ff-NLL.zigbee",
        "minFileVersion": 458752
    }
]
// minFileVersion: 458752 = 0x00070000 = "device must be on build 7 or higher"
```
A new PR to add the old v34 file gets this in the body:

```
[
    {
        "fileName": "1021-000f-00224203-NLL.zigbee",
        "maxFileVersion": 458751
    }
]
// maxFileVersion: 458751 = 0x0006FFFF = "device must be on build 6 or lower"
```

Would that work? 

Do I understand these file version numbers required correctly?

How do you force people in the future to add this json for new firmware version uploads?

Does z2m handle having two firmware versions for the same device correctly (v71 and v34)? Alternatively we could only add the new firmware version with the minFileVersion/json - users take care of updating old devices using the custom OTA function, say if they have < v7 in this case). z2m would just not offer an update if they are below 7 if I understand this correctly.

Or just trust that there are not too many old devices around, so far I haven't found a trace of trouble due to this "in-between" requirement online.

alexhk:
Probably worth remembering this behavior here:
https://github.com/Koenkk/zigbee-OTA/pull/1171#issuecomment-4631538278

 (z2m won't offer OTA update to newer version if two files are present in index)

PR #748, opened 2025-04-17
Add FW 1.1.2 for _TZ3000_w0qqde0g

mrmaximas:
```json
{
  "manufacturerName": [
    "_TZ3000_w0qqde0g",
    "_TZ3000_dksbtrzs",
    "_TZ3000_fukaa7nc"
  ],
  "releaseNotes": "Optimize some logic of power detection and add overcharge protection function"
}
```

PR #752, opened 2025-04-17
fix: `manufacturerName` mismatch when none provided in PR

Nerivec:
Should merge at least before https://github.com/Koenkk/zigbee-OTA/pull/748

@Koenkk I think that's the proper way? If no manuf name given in PR but existing has manuf name, it should not override it.


mrmaximas:
@Nerivec thanks, I'll keep that in mind next time. I'm not (yet) experienced enough in creating PRs ;)

Koenkk:
Yes, thanks!

PR #763, opened 2025-05-07
Update sihas PMM300Z2 ota file to v8

shinasys:
{
  "releaseNotes": "Added cumulative production and CT (Current Transformer) settings. Reconfiguration is required after the OTA update."
}


Koenkk:
Thanks!

PR #764, opened 2025-05-07
Update sihas PMM300Z3 ota file to v9

shinasys:
{
  "releaseNotes": "Added cumulative production. Reconfiguration is required after the OTA update."
}


Koenkk:
Thanks!

PR #785, opened 2025-06-11
New OTA image for n1aeu1 (H1-EU-single) from Lumi/Aqara

fumantsu:
New OTA file for Smart Wall Switch H1 EU Single rocket


Koenkk:
thanks!

PR #807, opened 2025-07-10
Update hue.ts

### LINKED ISSUE #806: MIssing some Hue hardware versions
Looking through this list, I see some of my hardware versions missing. Can these be added, or were they intentionally skipped?

```
100b-104
100b-103
100b-100
100b-10c
```

https://github.com/Koenkk/zigbee-OTA/blob/2c7b60866e026960fe7a9b247a387ecfecd598f6/src/autodl/hue.ts#L18

Intecpsp:
Add missing device type ids, fixes #806

### DIFF
--- a/src/autodl/hue.ts
+++ b/src/autodl/hue.ts
@@ -16,6 +16,10 @@ type PageJson = {updates: ImageJson[]};
 const NAME = "Hue";
 const BASE_URL = "https://firmware.meethue.com/v1/checkupdate?version=0&deviceTypeId=";
 const DEVICE_TYPE_IDS: string[] = [
+    "100b-10c",
+    "100b-100",
+    "100b-103",
+    "100b-104",
     "100b-111",
     "100b-112",
     // '100b-113',


Nerivec:
All these return empty currently from what I'm seeing, so will be ignored (might want to comment them out like below entries that are known to not have any firmware available).

Intecpsp:
I saw they were empty too, but I thought I was doing something wrong. Why do you think they are returning nothing when they have [release notes](https://www.philips-hue.com/en-us/support/release-notes/lamps) less than a year old?

TheJulianJES:
Since they're not Bluetooth lights, those updates are not available via the Hue mobile firmware API.

Nerivec:
Ah right, I knew there was something, but couldn't remember what it was.

Intecpsp:
Ah, so this PR isn't needed then. I bet we don't know where to obtain these updates, do we?

TheJulianJES:
The images in the repo will (manually) be updated soon, but the repo already has the latest ones for the IDs you added in this PR.

PR #812, opened 2025-07-13
Manual fixes for Hue OTAs

TheJulianJES:
Follow-up to:
- https://github.com/Koenkk/zigbee-OTA/pull/811

This adds min/max file versions for the added Hue OTAs.

---

Semi related, but the order for the metadata keys does not seem to be fixed (so it's not always the same).
My scripts always assumed the following order for the OTA metadata. However, some entries have the `originalUrl` in a different place (after `fileSize`).
Technically, it doesn't matter of course and I've updated my scripts to keep whatever order is present, but maybe the repo scripts should be updated to consistently follow one order?

Order I assumed to be standard (and it seems to be be that way for most entries):
- `fileName`
- `fileVersion`
- `fileVersion`
- `fileSize`
- `url`
- `imageType`
- `manufacturerCode`
- `sha512`
- `otaHeaderString`
- `originalUrl` (optional)
- `maxFileVersion` (optional)
- `minFileVersion` (optional)

### DIFF
--- a/index.json
+++ b/index.json
@@ -659,7 +659,8 @@
     "sha512": "4d15669f586c39da05fdfa95506fa085e297e177e5f2e962ce5f8ec2788f1d4488f880c35c2f2a1e0279ffe66272c99d27ca6da5d80512b88fcf50a574647a64",
     "otaHeaderString": "",
     "originalUrl": "https://otau.meethue.com/storage/ZGB_100B_0110/77cabadf-422e-4b65-a971-b40f6422ca6f/100B-0110-01002602-ConfLight-Lamps-EFR32MG13.zigbee",
-    "minFileVersion": 16778240
+    "minFileVersion": 16778240,
+    "maxFileVersion": 16786945
   },
   {
     "fileName": "100B-0111-01001D00-ConfLight-ModuLum-EFR32MG13.zigbee",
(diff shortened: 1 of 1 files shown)


index.json
```diff
@@ -978,7 +986,7 @@
     "manufacturerCode": 4107,
     "sha512": "a2a0076275a1dcdb94a9bea8c1591209f31f76fe847d5fd46456dad63247d0c302e438776c179dcc86f93471b3965e884f3455c06c332313fdc408df71a58c11",
     "otaHeaderString": "",
-    "originalUrl": "https://firmware.meethue.com/storage/100b-125/33571585/17803020-a40e-4015-b9e3-3454e43998bb/100B-0125-02004301-ContactSensor-EFR32MG22.zigbee",
+    "originalUrl": "https://otau.meethue.com/storage/ZGB_100B_0125/9d8ae429-ad80-434b-8e18-cc969b2f34d5/100B-0125-02004301-ContactSensor-EFR32MG22.zigbee",
```
TheJulianJES, line 989: This image was previously grabbed from the mobile API via the repo auto-download and it's the only image from `firmware.meethue.com`. However, it's also available from `otau.meethue.com`, like all other images, so that's also changed here. It's the exact same file (same `sha512`).
Technically, this change is not needed and could be taken out if we don't want to change it. I would though.

Nerivec: Should be fine. The next autodl should recognize the image as identical and ignore it. However, if a future autodl sees a newer image, it will update it.

Nerivec:
The order indeed does not matter, and was made totally arbitrarily from how the code was written. Between old entries, extra metadata provided through PR body and reparsing logic, there are a lot of ways the order can get shuffled (keys are appended in whatever way they come in). It's mostly machine-maintained & machine-used, so, cleaner to not add extra code for that (and faster).

Koenkk:
Thanks!

PR #813, opened 2025-07-14
Add latest Develco firmwares

samuli-as:
Add the latest firmware files for Develco devices (from https://github.com/Develco-Products/frient_upgrade_images).


Koenkk:
Thanks!

PR #833, opened 2025-08-14
add firmware version 67 for wb-msw-zigbee

AndreyKsenofontov:
Fixed firmware update if the main chip is on startup in the bootloader.


Koenkk:
Thanks!

PR #856, opened 2025-09-07
Add `modelId` for Bosch devices that didn't had it yet

DerDreschner:
Within the last weeks, the now available Bosch firmware files were added to the repository. During that process, not all entries got the corresponding `modelId` attribute. This pull request adds it as I'm not sure when it's necessary and when not (e.g. the relay and dimmer updated fine without it).


Koenkk:
If Bosch implements the spec properly, this shouldn't be needed. (I guess it's not needed)

DerDreschner:
Yeah, that's not necessary. I honestly have no idea why I've added it previously on some Bosch OTA updates. :thinking: Thought you told me to do so, but after some research, that's not the case...

Anyway, closed.

PR #857, opened 2025-09-08
Revert "update thirdreality smart plug ota file to v96"

weihuan1111:
Reverts Koenkk/zigbee-OTA#852


weihuan1111:
@Koenkk our smart plug ota file v96 have some questions, can you help to revert this version, and keep the latest ota file as v92, thank you.

PR #873, opened 2025-09-22
Bosch Room Thermostat II (`BTH-RM`): Add firmware v0.2.9 from April 2nd, 2025

DerDreschner:
Bosch released a firmware update for the Bosch `BTH-RM` (Room-Thermostat II) in [April 2025](https://www.bosch-smarthome.com/uk/en/support/release-information/system-solutions/2025-04-02.html). This update isn't included in the [official Bosch firmware update repository](https://github.com/BoschSmartHome/bosch-smart-home-files), which is the reason it's not included here, yet. It doesn't have a lot of changes, so, it's not that important. But as I was able to capture it, I add it here anyway.

```json
{
    "releaseNotes": "New bootloader"
}
```


Koenkk:
Thanks!

burmistrzak:
@Koenkk Unfortunately, this firmware seems to cause excessive battery consumption, as reported [here](https://community.bosch-smarthome.com/t5/technische-probleme/raumthermostat-2-batterie-alle-2-monate-leer/td-p/93359).
While the official OTA hasn't been pulled (yet), we should consider doing so or at least add a warning.

Koenkk:
Or should we revert this?

burmistrzak:
> Or should we revert this?

@Koenkk It's tricky because Bosch have to some extent acknowledged the issue, but haven't themselves pulled the update!? I also don't have a device with older firmware to develop against... 😬
A warning might be the best compromise, IMHO.

Koenkk:
Could you make a PR to update the device docs?

PR #879, opened 2025-09-26
Add OTA Legrand firmware file

OliviezGit:
Mirror Legrand firmware release :  24 septembre 2025

NLP device firmware, not mirror from legrand site, don't know if legrand downgrade or not this firmware... may be a legrand release issue...


Koenkk:
Thanks!

PR #883, opened 2025-10-03
add firmware version 70 for wb-msw-zigbee

AndreyKsenofontov:
Fix for the motion sensor – it could occasionally report motion detection when there was none.


Koenkk:
Thanks!

PR #895, opened 2025-10-14
OTA for lumi.switch v40

nickgoh:
This OTA firmware was extracted directly from an Aqara Gateway during a live firmware update.

Compatibility:
lumi.switch.acn055
lumi.switch.acn054
lumi.switch.acn049
lumi.switch.acn048

Notes:
Verified and tested with Zigbee2MQTT local OTA update for all four models. All devices updated successfully and are functioning as expected.


Koenkk:
Thanks!

PR #902, opened 2025-10-19
add firmware version 71 for wb-msw-zigbee

BurchuMikhail:
Fix for sensor wb-msw-zigbee - sensor values ​​were not sent.


Koenkk:
thanks

PR #913, opened 2025-11-06
Delete images/Candeo/C-ZB-LC20_V30151203_20250904.ota

candeodevelopment:
Temporarily remove OTA file for Candeo C-ZB-LC20 devices to allow investigation of reported OTA version anomalies.

We'll update with a new OTA file hipefully in the next few days.


Koenkk:
thanks!

PR #917, opened 2025-11-11
Create C-ZB-LC20v2_V30251203_20251110.ota

candeodevelopment:
@Koenkk this is the new OTA file for the Candeo LC20 series of devices which was temporarily removed in https://github.com/Koenkk/zigbee-OTA/pull/913

To avoid conflicts, ould you make sure that it is set to only apply to devices with the manufacturer name of "Candeo" please?

```json
{
    "manufacturerName": ["Candeo"]
}
```


Koenkk:
Done, thanks!

PR #932, opened 2025-11-28
sync the minFileVersion for 6 hue image types

dphi:
Hi, this PR syncs the minFileVersion for the following image entries that are _almost_ the same except for the url parameter:

manufacturerCode: 4107, imageType: 274, fileVersion: 16788226
manufacturerCode: 4107, imageType: 276, fileVersion: 16787202
manufacturerCode: 4107, imageType: 277, fileVersion: 16782594
manufacturerCode: 4107, imageType: 282, fileVersion: 16781314
manufacturerCode: 4107, imageType: 285, fileVersion: 16787202
manufacturerCode: 4107, imageType: 286, fileVersion: 16786690

I wanted to update my 4107 / 285 lamps but they always wanted to update to the latest version which was missing an intermediate step, likely due to the missing minFileVersion on one of the two alternatives for the latest version. - I have to say likely because I solved it for myself by upgrading to another intermediate version with an index.json that did not contain the images with the file versions this PR is updating. 

An alternative to this PR would be to remove the redundant entries.

### DIFF
--- a/index.json
+++ b/index.json
@@ -7677,7 +7677,8 @@
     "imageType": 274,
     "manufacturerCode": 4107,
     "sha512": "69f366711806331acc0dbc0a944f193feff9494c39854e4394bfdd6923a6b00698238c8046ab4843e65279365a7058a585126c47cac4bdc36d040d4a28f4eb23",
-    "otaHeaderString": ""
+    "otaHeaderString": "",
+    "minFileVersion": 16787714
   },
   {
     "fileName": "100B-0114-01002702-ConfLightBLE-Lamps-EFR32MG21.zigbee",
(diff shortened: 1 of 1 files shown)


TheJulianJES:
The auto-download script will always break Hue OTAs at the moment, as it ignores intermediate steps completely.
That's why I was not a fan of that being added for Hue. It's a pain to always fix them when properly adding the latest OTAs from the correct Bridge API, and even more so when they "randomly" break because some new update without min/max versions was added from the mobile OTA API.

My opinion is still that the auto-download for Hue should be removed/disabled entirely. There should be no Hue OTAs pointing to `firmware.meethue.com` (Mobile API), only to `otau.meethue.com` (Bridge API). The latter of which all have/had the correct min/max versions for all Hue lights requiring intermediate updates when they were added.

Koenkk:
@TheJulianJES would there be a way to integrate this in the script?

PR #940, opened 2025-12-03
Added the v1.0.8 firmware of ZBMINIR2

1775939187:
Add binding function: The device can only perform binding operations after the relay disconnect mode is enabled.


Koenkk:
Thanks!

andrei-lazarov:
Awesome!! 😁 Thank you ✌🏻 
Fixes: https://github.com/Koenkk/zigbee2mqtt/issues/25392

uschindler:
Awesome! There's also another change with the firmware update: In "pulse trigger" mode, the device now switches relay when button is pressed. Previously it changed the relay state when button was released. This makes the buttons react instantly and there's no longer a delay.

But in general it is good that you can now bind lamps in detached relay mode. I have not yet tested this, but that's promising. I hope sonoff will also apply the same the SNZB buttons, which can't be bound to groups yet.

andrei-lazarov:
@uschindler 

I can bind my **SONOFF SNZB-01** to individual devices and/or groups. Hard-coded behavior:
- single_press: TOGGLE (individually toggle each device / member of group)
- double_press: ON
- long_press: OFF

The **SONOFF SNZB-01P** does not have binding. 
They tried to make it work: https://github.com/Koenkk/zigbee2mqtt/issues/23304

But it missing binding is currently a selling point for the new **SONOFF SNZB-01M**, which they claim has binding.

sonoff_buttons

PR #958, opened 2025-12-15
Update 3RSPU01080Z

weihuan1111:
```json
{
    "modelId": "3RSPU01080Z"
}
```


weihuan1111:
Hi, @Koenkk in pull request https://github.com/Koenkk/zigbee-OTA/pull/953 ,we don't need to add the manufacturerName, only need to add modelId, and we have tested that remove manufacturerName can OTA success, so, can you help to remove the manufacturerName

PR #970, opened 2025-12-17
Remove old pvvx custom firmware

web2brain:
Newer version has been added with https://github.com/Koenkk/zigbee-OTA/pull/949

### DIFF
--- a/index.json
+++ b/index.json
@@ -5360,27 +5360,6 @@
     "otaHeaderString": ""
   },
   {
-    "fileName": "1141-0208-01143001-ZMHOC401N.zigbee",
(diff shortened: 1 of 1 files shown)


web2brain:
It seems like removing old ota files that have a more recent version in another subdirectory can't be removed as easily? Do I need to edit the index file manually? I doubt that.

web2brain:
@Koenkk I wanted to add the label `ignore-ota-workflow` but unfortunately can't, probably I don't have the rights. Or is there another way to remove the old images than editing the `index.json` manually?

Koenkk:
Thanks!

PR #971, opened 2025-12-18
Added Ota files for SONOFF device ZBM5 

1775939187:
“zbm5-120-zr” add binding function: The device can only perform binding operations after the relay disconnect mode is enabled.


Koenkk:
Thanks!

PR #978, opened 2025-12-25
Added the v.1.0.3 OTA image for SNZB-02DR2

Qiu-A-Qiu:
Fix the display bug of 02DR2 temperature higher than 99 and lower than 0


Koenkk:
Thanks

PR #984, opened 2025-12-29
fix: dedupe by manufacturer

Nerivec:
- cleanup Hue images (no dupe, all versioned in base) Closes #932
- process all images in Hue autodl instead of latest to add `minFileVersion` meta
- prevent SHA-matching or spec-matching images from being added for same manufacturer
- `manufacturerName` matching on any (not every)
- add a few new Hue images (manual autodl run)
- add couple of scripts for local testing/checking (spec matching, and Z2M matching)
- fix biome related issues


Koenkk:
thanks

Issue #972, opened 2025-12-18
Aqara OTA updates


ditheo:
I cannot ota update AQARA devices such as aqara WKS-EU03 and WKS-EU04.

I can see on the repository there is a new firmware with date  26/07/2024 and version 25 for WKS-EU03 (lumi.switch.n2aeu1) but if I try to check for updates i cannot see the new firmware. My devices are on version 21/10/2021 and file vesrion 2839.

I cannot see WKS-EU04 on the OTA update section, even the device supports OTA updates.

I have the 2.7.1 version z2m and for adapter I use a smlight ZB-06M

Koenkk:
We don't have the newer OTAs, to retrieve it from the Aqara hub see https://www.zigbee2mqtt.io/advanced/more/tuya_xiaomi_ota_url.html

ditheo:
The version file that I wrote ( previous and newer) are from the zigbee ota collection mentioned on index.json and they exist under folder images/lumi. I think that is the way that someone can do ota update to zigbee devices, If I have misunderstanding that please don't bother

andrei-lazarov:
Hi @ditheo! Thanks for spotting the issues. I'll recap:

Devices are: 
1. 1-gang, `lumi.switch.n1aeu1`, WS-EUK03, imageType 5404, OTA enabled
2. 2-gang, `lumi.switch.n2aeu1`, WS-EUK04, imageType 5532, **OTA not enabled** (we should enable this)

There is a 2021 update for 1-gang and 2024 updates available for both ([release](https://github.com/Koenkk/zigbee-OTA/releases/tag/2025-06-15T021848021Z)): 
- 1-gang 2021 version 2839
- 1-gang 2024 version 25
- 2-gang 2024 version 25

**The fileVersion is lower for some reason.**
- We can skip the version check and perform the update with `force: true`.  
But then, the update will keep showing up, even after it was installed. (until we remove `force:true`)

**You can follow the instructions in #786 to update:**
Enable OTA with custom converter, add custom index with `force`, update, remove custom index (`force`).

---

**But we should find a solution, so other users don't have to manually do the steps above.**
@Koenkk 
> I'm wondering if we can trick it by increasing the `"fileVersion": 25`
- If we just increment `fileVersion`, Z2M will refuse the update, 
because it checks the header inside the OTA file.
- So we must also hex-edit the header and update the hash. 
The `fileVersion` is stored as 4 little-endian bytes, at offset 14:
[...]

PR #986, opened 2025-12-29
fix: aqara/lumi enforce `modelId` meta

Nerivec:
Cleaned up files that were added without proper metadata and bypassed the checks.
cf: https://github.com/Koenkk/zigbee-OTA/issues/972#issuecomment-3678355701


Koenkk:
Thanks!

PR #1006, opened 2026-01-07
Restore deleted images

andrei-lazarov:
- Fixes https://github.com/Koenkk/zigbee-OTA/issues/995

Added the following updates to `images-old/`:
- IKEA motion sensor E1745 v24.4.5
  - https://github.com/Koenkk/zigbee2mqtt/issues/25587
- Aqara smoke sensor v21
  - https://github.com/Koenkk/zigbee-OTA/issues/938 
  (alarm stops after 1 beep)
- SONOFF SNZB-06P v1.0.6
  - https://github.com/Koenkk/zigbee2mqtt/issues/22071#issuecomment-3575643493
- SONOFF ZBMINIR2 v1.0.4
  - https://github.com/Koenkk/zigbee-OTA/pull/985#issuecomment-3708659122

PR #1008, opened 2026-01-08
📝 Document images-old

andrei-lazarov:
- https://github.com/Koenkk/zigbee-OTA/pull/1006#issuecomment-3720365815

### DIFF
--- a/images-old/README.md
+++ b/images-old/README.md
@@ -1,0 +1,33 @@
+## Removed images
+
+### IKEA
+
+- **E1745 TRADFRI motion sensor v24.4.5**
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Nerivec:
I think it's best to rename the folder more clearly, with the new "easy" OTA upload capability, I'm a little worried with this...
Maybe something like `RETRACTED_IMAGES` (so users are pushed to look at the readme, "old" doesn't sound like a warning)?

andrei-lazarov:
Yes, retracted sounds good!

Nerivec:
Done, also added a big caution sign in the readme :grin:

PR #1036, opened 2026-01-22
New image for lumi.switch.acn040 device from Lumi

macher91:
This is the latest OTA file for device lumi.switch.acn040.

```json
{
    "modelId": "ZNQBKG31LM",
    "manufacturerName": ["Aqara"]
}
```


andrei-lazarov:
Hi. I believe modelId should be `lumi.switch.acn040`.
Also did you try the update? Is it stable?

klimeryk:
> Hi. I believe modelId should be `lumi.switch.acn040`.

Yup, that seems to be the case:

Screenshot 2026-04-01 at 22 48 58

> Also did you try the update? Is it stable?

I'm not the author of the PR, but I stumbled upon it while trying to fix the issue discussed in https://github.com/Koenkk/zigbee2mqtt/discussions/25423#discussioncomment-15795254 (TL;DR: the switch binds on Zigbee protocol level with a random IKEA remote...). And indeed, after applying this update, the issue went away 🎉 

As can be seen on the screenshot above, the new firmware (from this PR) is reported as `Firmware version: 29`, while the old one is `27`.

I actually have two of these switches, I've left the other one on the old firmware in case you need any extra info or testing. So far everything seems to work as before, no regressions, no difference in features. I've been running the new firmware for about a week.

Let me know if I can provide any other information 🙇

andrei-lazarov:
All good, thanks @klimeryk! I'll start a new PR to replace this one.

PR #1038, opened 2026-01-24
fix: cleanup `customParseLogic` no longer needed

Nerivec:
No longer needed after refactor: https://github.com/Koenkk/zigbee-herdsman/issues/1608 (parsed properly, automatically, based on tag ID)


Koenkk:
Thanks!

PR #1058, opened 2026-02-04
add firmware version 72 for wb-msw-zigbee

BurchuMikhail:
Added automatic baud rate detection on first launch and connection reset.


Koenkk:
Thanks!

PR #1073, opened 2026-02-19
Cleansing firmware folder pvvx from old orphans

yannpub:
Removing old orphans from folders images and images1, leaving only 

- the latest version v133 in folder images (21 firmware files)
- and v132 in images1 (21 firmware files)


yannpub:
@Koenkk : Should I only delete the orphans from folder images? Will it then automatically remove the version minus one from the images1 folder?

Koenkk:
> Will it then automatically remove the version minus one from the images1 folder?

yes those should be automatically removed, so no need to manually remove images.

yannpub:
Will re-do another PR, only deleting from folder images, without any change on folder images1.

PR #1074, opened 2026-02-19
Cleansing of old pvvx firmwares

yannpub:
Removing orphan fiirmware files

### DIFF
--- a/index.json
+++ b/index.json
@@ -5521,17 +5521,6 @@
     "modelId": "4512760"
   },
   {
-    "fileName": "1141-0201-01263001-ZMHOC401.zigbee",
-    "fileVersion": 19279873,
-    "fileSize": 132130,
-    "url": "https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/pvvx/1141-0201-01263001-ZMHOC401.zigbee",
-    "imageType": 513,
-    "manufacturerCode": 4417,
-    "sha512": "154985074600f19f5f470a70d36056a8665a06adc68a6ccb49ab7ccb822fffa570271349f79f8f4911dd73642933449fb931b2c901f2facc17b07e759be2c874",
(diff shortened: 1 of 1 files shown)


yannpub:
@Koenkk : Here are the firmwares files that are outdated in the folder images/pvvx. 
The workflow is failing as it is deletion of files --> Is it a case the PR should be labeled "ignore-ota-workflow" ? 

f I understand well, the index.json will be automatically fixed as it will not find the entries anymore in the folder, right ? Or will I need to manually update the index.json once the files are deleted?

Koenkk:
>  Or will I need to manually update the index.json once the files are deleted?

These need to be manually deleted, why have these images become orphan?

yannpub:
This is a good question for which I do not have an answer ;-) My guess would be that it happened when there was a change in the script that made new fields mandatory, it did not delete the previous ones and considered the upgraded firmware as new ones ?

We should only have the v133 in folder images and the v132 in the downgrade folder. This is why I wanted to remove the other ones as keeping them does not make sense.

Koenkk:
I see, then also update `index.json`.

kwiateusz:
@yannpub Hi latest version is 0.1.3.4 can you add this while you're at it?

yannpub:
@Koenkk : Index.json updated with the removal of the orphans firmwares. Script still prevent the PR to be validated. 

@kwiateusz : V0.1.3.4 hasn't been released 'officially' by Victor. I only update here once Victor has done a release.

kwiateusz:
Even if the version is already in the readme and in the bin directory, it is not an official release?

Koenkk:
Thanks!

yannpub:
@kwiateusz : it is not. There was still some commits yesterday. I only upload when it is finalised and released.

PR #1094, opened 2026-03-09
Fix Aqara WS-EUK03 and WS-EUK03 (lumi.switch.n1aeu1, n2aeu1) updates

andrei-lazarov:
Screenshot From 2026-03-10 00-20-11 (Edit)

Back to the Aqara updates with wrong file version. cc @Nerivec 

Indeed, after updating, the device reports the correct file version. 
old: 2839 -> new: 2841, not 25.

Only the OTA header is wrong. It's missing the `b` in this case.

So I edited the files manually to match.

- fixes https://github.com/Koenkk/zigbee2mqtt/issues/31318


Nerivec:
Should not edit binary files, this repo is just a mirror.
End users can if they wish, at their own risk...

PR #1100, opened 2026-03-16
Add firmware version 73 for wb-msw-zigbee

BurchuMikhail:
The sensor's microcontroller firmware has been updated.
An issue with the sensor freezing after repeated power applications (more than 200) has been resolved.


Koenkk:
Thanks!

PR #1103, opened 2026-03-18
add firmware version 75 for wb-msw-zigbee

BurchuMikhail:
1. Fixed a connection termination error
2. Add a sensor reflashing function in case of failures


Koenkk:
thanks!

PR #1141, opened 2026-04-15
add ota for home-versions of ONOKOM devices

zwg3:


zwg3:
@Koenkk 
Hello, sorry to bother you.
It seems that we can't pass the necessary checks to complete the PR process. The issue, it seems, is that the added files have manufacturerCode, imageType and fileVersion fields that match those of the existing files. The problem here is that, although these OTAs are based on the same protocols, these devices do have different functionality at the cluster level and therefore should be separated from one another.
So, although there already exists a file like aux-1-zb-s_hw1.x_fw0.6.1_ota20, how should we go about adding the ok-ac-h-aux-1-zb-s-a_hw1.x_fw0.6.1_ota20 file that would have the same manufacturerCode, imageType and fileVersion fields?

Koenkk:
> how should we go about adding the ok-ac-h-aux-1-zb-s-a_hw1.x_fw0.6.1_ota20 file that would have the same manufacturerCode, imageType and fileVersion fields?

This should not be done as z2m cannot decide what image to upload to what device, a different device should have a different `imageType`.

PR #1147, opened 2026-04-19
Add Develco/Frient OTA images

samuli-as:
Added latest firmware images from **https://github.com/Develco-Products/frient_upgrade_images**


samuli-as:
Seems two of the images were duplicates as some devices share the same image. Removed.

Koenkk:
Thanks!

PR #1167, opened 2026-05-11
Cleansing - Old pvvx firmwares

yannpub:
Removing old orphans in order to keep only latest versions (v135 in images1 and v136 in images)


Koenkk:
thanks!

PR #1171, opened 2026-05-15
Legrand 067776A Firmware v71 May 15 2026

alexhk:
New firmware 71 for Legrand 067776A Netatmo wired shutter switch with level control (NLLV). Original file download captured via tcpdump/traffic of Legrand Gateway. Tested via manual z2m OTA update to some of my devices.

Edit:

minFileVersion: 458752 = 0x00070000 = "device must be on build 7 or higher"

```json
[
    {
        "fileName": "1021-000f-004745ff-NLL.zigbee",
        "minFileVersion": 458752
    }
]
```


andrei-lazarov:
Hi. Can you check (and revive) this thread?
- https://github.com/Koenkk/zigbee-OTA/pull/751

I see the updates are available for download after you create an account.
But for a few devices you can't update directly to the latest version, there is one update in-between

alexhk:
@andrei-lazarov 

I saw that thread and I know the device firmware download page. Let's just say it's complicated.

The 067776A uses the NLL firmware. On that page, it is served for the Light Switch with Neutral. While it's not the fw 71, and if I remember correctly, I compared the checksum of the available download fw 67 with the firmware z2m has for the shutter, it's the same file.
On the other hand, they list the Shutter Switch with Neutral as NLV type, with another firmware file. I believe that is the 067776 non-A (old version). The new 067776A they don't even list as variant on that page. So I assume it's outdated.

image


I tried to make sense of it for a few days and gave up. As far as I can tell Legrand doesn't keep that download page up-to-date. There are fw version in production (served by Legrand Gateway to devices) ahead of what is on that developer download page, I think in the thread this was also mentioned.

It also gets more confusing because the device type names and firmware names. Case in point:
NLL family: imageType 0x000f, covers multiple devices including 067776A, 067797, etc.
NLV family: imageType 0x0013, used by the original 067776 non-A variant, which has no level control.
NLLV: Legrand's hardware/product designation for the 067776A's electronics/component generation. NLLV devices use NLL firmware binaries (per Legrand dev forum), not a separate NLLV-named firmware. And yes, there are also NLL type devices without the V. It's a mess. If you look at  Image Block Requests to the Legrand Gateway in Wireshark it gets even more confusing, as the application builds don't seem to match.

Also look at this (no account required):
https://developer.legrand.com/local-interoperability/#How%20it%20works
They are listed as NLLV there it seems. The old ones (NLV) are not even in the same list it seems.
Shutter switch with neutral	SHUTTERS	NLLV
Light switch with neutral	LIGHTING	NLL

Update in-between old versions:
Legrand's dev download page lists for NLL type firmware that if your fw is <7 you need to use fw 31 first, before updating. I am not sure when fw 7 happened and I have only used Legrand devices for a few months.  Does z2m have such conditional updates for any of the other devices?

alexhk:
@andrei-lazarov 

I changed this to draft.

Investigating if we should add something like this (if this can even be done, otherwise new PR it is if it's confirmed to be needed):

```
[
    {
        "fileName": "1021-000f-004745ff-NLL.zigbee",
        "minFileVersion": 458752
    }
]
// minFileVersion: 458752 = 0x00070000 = "device must be on build 7 or higher"
```

andrei-lazarov:
Yeah it seems complicated. Thanks for investigating!
It's ok to add this device only, it's up to you.

Yes minFileVersion + multiple images should work. I think it's used for Philips Hue.
You can edit the initial message in this PR and the bot will run again.

Koenkk:
ready for merge?

alexhk:
Yes, please go ahead.

Koenkk:
Thanks!

alexhk:
@andrei-lazarov I think the back and forth we had here resulted in the commit not removing the old entry, maybe because the old entry didn't have a min or max version specified.

As a result we have now both entries in the index and no OTA update is offered via z2m:

```
Old v67 firmware:

  {
    "fileName": "1021-000f-004345ff-NLL.zigbee",
    "fileVersion": 4408831,
    "fileSize": 255127,
    "url": "https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Legrand/1021-000f-004345ff-NLL.zigbee",
    "imageType": 15,
    "manufacturerCode": 4129,
    "sha512": "aa89a06714e446662dc2f183f0cd72a54dcf344ed62a30ab14b12943eca06857418eba4958398452af60470dc9665e380f155f7d693abb79532d55759b75d772",
    "otaHeaderString": "                                "
  },
   
New v71 firmware:

{
    "fileName": "1021-000f-004745ff-NLL.zigbee",
    "fileVersion": 4670975,
    "fileSize": 255143,
    "url": "https://raw.githubusercontent.com/Koenkk/zigbee-OTA/master/images/Legrand/1021-000f-004745ff-NLL.zigbee",
    "imageType": 15,
    "manufacturerCode": 4129,
    "sha512": "dc9bb4669f947be422747de971285f917301b343bb1be5e78e7dbbe68988ea1fcaf1eaa0989a22a2f64647f0897c1d45d7967e08f36d2c4c8634a8fd9581efd4",
    "otaHeaderString": "                                ",
    "minFileVersion": 458752
  },

```

andrei-lazarov:
What is the version currently installed? Did you make sure the imageType and manufacturerCode match? (see debug logs when checking for updates)

alexhk:
Installed is v67. z2m should offer the update for v71.

Currently in the index we have v67 and v71. When I reverse the order (v71 comes first) in the index.json z2m finds a news OTA update, even with both present. But in the order v67 and then v71 like it is now, it doesn't find a new update. And of course if I remove the v67 and only keep v71, it also offers and update.


Should the github-action bot have removed v67 from the index or kept it?

At the beginning the github-action bot wanted to remove the old and replace it with new:
https://github.com/Koenkk/zigbee-OTA/pull/1171#issuecomment-4462225920

Later, presumably after adding `"minFileVersion": 458752`, it only added the new one, didn't remove the v67:
https://github.com/Koenkk/zigbee-OTA/pull/1171#issuecomment-4467090390

PR #1197, opened 2026-06-15
Feat: add Heiman HS1RM-E ota firmware

Leo2442926161:
What's new:
1. Add switch type selection.
2. Add switch power on behavior selection


Koenkk:
Thanks!

PR #1243, opened 2026-07-15
Bosch BTH-RA/-RAD (Radiator Thermostat II / II [+M]): Add updated firmware files from July 14th, 2026

DerDreschner:
This week, Bosch released a new firmware for their Radiator Thermostat II models:

- Radiator Thermostat II (`BTH-RA`): version 3.08.03
- Radiator Thermostat II [+M] (`BTH-RAD`): version 1.03.09

They were [pretty quick](https://github.com/BoschSmartHome/bosch-smart-home-files/commit/aaadcfc534c68fb0c9eed3294b8a1f80c9e24de9) with releasing them in their public GitHub repository, where I've pulled the files in this PR from. According to [their change log](https://www.bosch-smarthome.com/uk/en/support/release-information/system-solutions/2026-07-14.html), they changed the minimum valve position to 5% in low power situations.

Unfortunately, the firmware for the Room Thermostat II 230V [still isn't included](https://github.com/BoschSmartHome/bosch-smart-home-files/issues/2) in their repository. I'm sure it's very much appriciated by the community if anyone with the device is able to sniff the update.

```json
[
    {
        "fileName": "TRV_GEN2_0x38031514_PROD.fw",
        "releaseNotes": "The value for the valve opening when the battery is very low was reduced to 5% to prevent overheating when the battery is flat and to still guarantee frost protection."
    },
    {
        "fileName": "TRV_GEN2_DUAL_0x13098210_PROD.fw",
        "releaseNotes": "The value for the valve opening when the battery is very low was reduced to 5% to prevent overheating when the battery is flat and to still guarantee frost protection."
    }
]
```

PR #1245, opened 2026-07-16
fix: Add `modelId` for Bosch BTH-RA/-RAD

DerDreschner:
My previous PR #1243 didn't include the `modelId` for the new files. That made the GitHub Action not moving the old firmware files into the `images1` folder as well as the devices not discover the new firmware. This PR adds the model IDs and I hope that's enough to make the workflow archiving the old versions.


Koenkk:
Thanks!

PR #1266, opened 2026-08-08
Legrand production firmware july 30, 2026

OliviezGit:
Last update July 30, 2026


OliviezGit:
Still not original NLL firmware already published here...
This one is coming from legrand production firmware so it is the original one...

PR #1271, opened 2026-08-13
Add SONOFF SWV-ZNE/ZFE/ZNU/ZFU v1.1.0 OTA image

CubeZ2mDeveloper:
## Summary

Add a new OTA image for `SONOFF`  `SWV-ZNE` /  `SWV-ZFE` /  `SWV-ZNU` / `SWV-ZFU`.

## Device

- Product: SONOFF SWV-ZNE / SWV-ZFE / SWV-ZNU / SWV-ZFU
- Firmware version: v1.1.0

## Changes

- Added support for Imperial gallons.
- Fixed several RTC and battery reporting issues.
- Improved Bluetooth connection reliability by automatically disconnecting after prolonged inactivity.

## Testing

Tested on:
- Device: `SONOFF SWV-ZFE`
- From version: `v1.0.9`
- To version: `v1.1.0`
- Dongle: `SONOFF ZBDongle-E`

## Notes

- No breaking changes.
- Official vendor firmware.

```json
{
    "releaseNotes": "- Added support for Imperial gallons.\n- Fixed several RTC and battery reporting issues.\n- Improved Bluetooth connection reliability by automatically disconnecting after prolonged inactivity."
}
```

PR #1273, opened 2026-08-14
Revert "Add SONOFF SWV-ZNE/ZFE/ZNU/ZFU v1.1.0 OTA image"

CubeZ2mDeveloper:
## Reason

Revert #1271 because the OTA image was detected by devices other than the intended models, including the SONOFF MINI-ZBRBS.

Further investigation is needed to verify the OTA matching conditions and ensure the update is only offered to the intended devices.


CubeZ2mDeveloper:
Hi @Koenkk, the check for this PR is failing because the automatic revert also removed the OTA image file:

`[images/Sonoff/SN-TLSR8656-SWV1C-01-v1.1.0.ota] ENOENT: no such file or directory`

Could you please help revert/withdraw this PR? We will need to handle the OTA entry separately after further investigation of the matching issue.

Thanks!

Koenkk:
Reverted it!

PR #1291, opened 2026-09-01
fix(sonoff): add model IDs for SWV OTA firmware

CubeZ2mDeveloper:
## Summary

* Add separate OTA entries for `SWV-ZFE`, `SWV-ZFU`, `SWV-ZNE` and `SWV-ZNU`.
* Keep the OTA metadata identical for all four entries except for `modelId`.

## Motivation

The four SWV models use the same OTA firmware, but need to be distinguished by `modelId` to ensure the correct OTA firmware is detected.

Adding the model IDs directly to `index.json` also ensures compatibility with older Zigbee2MQTT versions.

As discussed in https://github.com/Koenkk/zigbee-herdsman-converters/pull/12960, this approach was confirmed to be acceptable.

### DIFF
--- a/index.json
+++ b/index.json
@@ -9954,6 +9954,42 @@
     "releaseNotes": "- Added support for Imperial gallons.\n- Fixed several RTC and battery reporting issues.\n- Improved Bluetooth connection reliability by automatically disconnecting after prolonged inactivity."
   },
   {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1303, opened 2026-09-15
Add new Innr firmwares

Koenkk:
Release notes:

> - Added support for color loop (Zigbee ColorLoopSet cmd) in all color lights
> - All color lights except ceiling lights now support color temperatures 1000K - 20000K
> - Improved color / color temperature accuracy for some lights
> - Improved white purity (planckian/daylight) for all color lights
> - Improved color accuracy at very low dim levels for luminaires with control box
> - Stepless dimming for OLS 210 with long press (factory reset now double-press/hold)
> - Fixed issue of tunable white lamp flash at specific dim level and color temperature
> - Several small fixes and stability improvements.

CC: @gerryinnr


uschindler:
No changes in "RB 286 C" lamp? It also is an RGB bulb, so I would have expected an update, too.

lwatsondev:
Just updated my smart plugs. I'm not entirely sure this is accurate. 🤨

Screenshot_20260916-194649

Pferdebockwurst:
What is. Ot 

> Just updated my smart plugs. I'm not entirely sure this is accurate. 🤨
> Screenshot_20260916-194649

Whats wrong with that?

lwatsondev:
> What is. Ot
> 
> > Just updated my smart plugs. I'm not entirely sure this is accurate. 🤨
> > Screenshot_20260916-194649
> 
> Whats wrong with that?

I'm fairly certain a domestic fridge does not use £346 of electricity in a single day.

Basically I updated my 242's firmware and about half of them out of 20 had their kWh readings inexplicably increased by about 1000 kWh. It's not a HASS bug as the same reading appears in Zigbee2MQTT, so firmware bug or z2m bug?

andrei-lazarov:
@lwatsondev  you can get the raw value directly from the device:
Go to Z2M -> device -> dev console -> read attributes -> seMetering: currentSummDelivered, divisor, multiplier

Z2M and HA shouldn't alter the values.
If the fw update really added 1000 kWh, go to Z2M -> device -> Settings (specific) and do -1000 offset in ~~Power~~ Energy calibration

If there's more to it, open a new issue to avoid spamming this thread

andrei-lazarov:
Some devices received updates, but OTA is not enabled. Submitted the PR above

AlexanderLeporiday:
The Innr SP 242 firmware has reverted a fix from previous firmware versions and now energy is reported 10x higher and requires a modification in the converter to reduce the values back down.

Likely this was changed because people on older versions on the converter raised issues but this is because the firmware had a bug and it was being corrected in converter.

Would be better if we could modifier the energy multiplier per device through the UI so we don't need an update to the converter every time this flip flops back and forth.

lwatsondev:
> Would be better if we could modifier the energy multiplier per device through the UI so we don't need an update to the converter every time this flip flops back and forth.

You kind of can. In Zigbee2MQTT, go to the device's "Settings (specific)" page and set "Energy calibration" to -90, assuming your exact x10 number is correct.

AlexanderLeporiday:
> > Would be better if we could modifier the energy multiplier per device through the UI so we don't need an update to the converter every time this flip flops back and forth.
> 
> You kind of can. In Zigbee2MQTT, go to the device's "Settings (specific)" page and set "Energy calibration" to -90, assuming your exact x10 number is correct.

That's perfect! I couldn't figure out what value to set here but -90 does appear to be accurate again. It's not exact due to rounding errors but it's much closer of an estimate.

AlexanderLeporiday:
Work around in Z2M is to go into each device and set the energy calibration to -90 and refresh the energy reading. This brings it back down to pre-firmware update numbers.

lwatsondev:
> Work around in Z2M is to go into each device and set the energy calibration to -90 and refresh the energy reading. This brings it back down to pre-firmware update numbers.

FYI some hardware revisions don't have 1.9.56 and don't need this fix. You only need to apply this to devices on 1.9.56, so check before you apply it to all devices. 👌

AlexanderLeporiday:
> > What is. Ot
> > > Just updated my smart plugs. I'm not entirely sure this is accurate. 🤨
> > > Screenshot_20260916-194649
> > 
> > 
> > Whats wrong with that?
> 
> I'm fairly certain a domestic fridge does not use £346 of electricity in a single day.
> 
> Basically I updated my 242's firmware and about half of them out of 20 had their kWh readings inexplicably increased by about 1000 kWh. It's not a HASS bug as the same reading appears in Zigbee2MQTT, so firmware bug or z2m bug?

It will be effecting all 242 (And likely all 24x) Innr devices due to a change in the firmware. We had this issue before where energy reported 10x higher than it should and this was fixed by Innr. The issue is people had coded around the firmware issue and suddenly saw 10x lower values than expected and older versions of Z2M were showing the wrong values. I suspect Innr received conflicting reports from users and reverted their change and now the latest version of Z2M shows the wrong values! Grr.

Solution is as outlined above. You have to go into each device and set a -90 calibration on energy. Until the fireware is fixed this is the solution. We should NOT update the Z2M converter back to a 0.1 scale unless Innr responds saying they will NOT be fixing the fireware bug otherwise we have to push out another update to the converter when the firmware is updated again!

GerryInnr:
> No changes in "RB 286 C" lamp? It also is an RGB bulb, so I would have expected an update, too.

Sorry this update is only for currently sold color lights.

GerryInnr:
> > > What is. Ot
> > > > Just updated my smart plugs. I'm not entirely sure this is accurate. 🤨
> > > > Screenshot_20260916-194649
> > > 
> > > 
> > > Whats wrong with that?
> > 
> > 
> > I'm fairly certain a domestic fridge does not use £346 of electricity in a single day.
> > Basically I updated my 242's firmware and about half of them out of 20 had their kWh readings inexplicably increased by about 1000 kWh. It's not a HASS bug as the same reading appears in Zigbee2MQTT, so firmware bug or z2m bug?
> 
> It will be effecting all 242 (And likely all 24x) Innr devices due to a change in the firmware. We had this issue before where energy reported 10x higher than it should and this was fixed by Innr. The issue is people had coded around the firmware issue and suddenly saw 10x lower values than expected and older versions of Z2M were showing the wrong values. I suspect Innr received conflicting reports from users and reverted their change and now the latest version of Z2M shows the wrong values! Grr.
> 
> Solution is as outlined above. You have to go into each device and set a -90 calibration on energy. Until the fireware is fixed this is the solution. We should NOT update the Z2M converter back to a 0.1 scale unless Innr responds saying they will NOT be fixing the fireware bug otherwise we have to push out another update to the converter when the firmware is updated again!

There was a regression in the last firmware updates for the SP 24x and OSP 240 smart plugs.
The issue is fixed and new firmware updates (versions 1.9.67 and 1.9.68) have been submitted for Z2M and ZHA.
Apologies for the inconvenience.

AlexanderLeporiday:
@GerryInnr Thanks for the fast turnaround on this one! Mistakes happen, it's great to see how quickly you and your team have responded to this issue and are getting it resolved. Now we just need to wait for @Koenkk to have some free time to push the new files out via OTA.

Pferdebockwurst:
> > > > What is. Ot
> > > > > Just updated my smart plugs. I'm not entirely sure this is accurate. 🤨
> > > > > Screenshot_20260916-194649
> > > > 
> > > > 
> > > > Whats wrong with that?
> > > 
> > > 
> > > I'm fairly certain a domestic fridge does not use £346 of electricity in a single day.
> > > Basically I updated my 242's firmware and about half of them out of 20 had their kWh readings inexplicably increased by about 1000 kWh. It's not a HASS bug as the same reading appears in Zigbee2MQTT, so firmware bug or z2m bug?
> > 
> > 
> > It will be effecting all 242 (And likely all 24x) Innr devices due to a change in the firmware. We had this issue before where energy reported 10x higher than it should and this was fixed by Innr. The issue is people had coded around the firmware issue and suddenly saw 10x lower values than expected and older versions of Z2M were showing the wrong values. I suspect Innr received conflicting reports from users and reverted their change and now the latest version of Z2M shows the wrong values! Grr.
> > Solution is as outlined above. You have to go into each device and set a -90 calibration on energy. Until the fireware is fixed this is the solution. We should NOT update the Z2M converter back to a 0.1 scale unless Innr responds saying they will NOT be fixing the fireware bug otherwise we have to push out another update to the converter when the firmware is updated again!
> 
> There was a regression in the last firmware updates for the SP 24x and OSP 240 smart plugs. The issue is fixed and new firmware updates (versions 1.9.67 and 1.9.68) have been submitted for Z2M and ZHA. Apologies for the inconvenience.

The OSP 240 is not enabled for OTA in Z2M.

AlexanderLeporiday:
> The OSP 240 is not enabled for OTA in Z2M.

As of 35 minutes ago, https://github.com/Koenkk/zigbee-herdsman-converters/pull/13214

This means that the next release of Z2M, with the next release of the converter will enable OTA for all Innr devices, however these devices also need to have firmware in the repo to actually update. But it will be easier to update them if firmware is missing once the OTA ability is pushed out.

PR #1304, opened 2026-09-16
Discover Inovelli firmware on GitHub

JamieMagee:
`firmware.json` lists VZM31-SN 3.04 as the latest release, so the autodl job misses newer files in Inovelli's GitHub repository. The updater now reads both sources and records each GitHub blob SHA to skip unchanged files.

The firmware update is in #1305.

Tests: `pnpm run build`, `pnpm test`, and `pnpm run check`.


JamieMagee:
Split the firmware files into #1305. This PR now contains only the autodl code, tests, and cache.

Koenkk:
Thanks!

rohankapoorcom:
@Koenkk I would suggest reverting this since this is pulling Inovelli's beta's and suggesting them for upgrade to the Z2M users. @InovelliUSA generally only wants platforms to suggest upgrades automatically after they are promoted from beta and then added to the stable OTA index file.

InovelliUSA:
Yeah, it looks like this new method grabbed a beta firmware. We have been maintaining the [firmware.json](https://[files.inovelli.com/firmware/firmware.json](https://files.inovelli.com/firmware/firmware.json)) file on our webserver to push out firmware to Zigbee2MQTT users.

JamieMagee:
Sorry, this was my mistake. I treated the GitHub repository as the source of truth and missed that `firmware.json` is Inovelli's curated feed for Zigbee2MQTT. I've opened #1310 to revert the autodl change and #1311 to restore the firmware state.

InovelliUSA:
Thanks!

PR #1310, opened 2026-09-18
Revert Inovelli GitHub firmware discovery

JamieMagee:
Reverts #1304. Inovelli maintains `firmware.json` as the curated Zigbee2MQTT feed. Scanning every OTA file in the firmware repository also selects beta releases, including VZM31-SN 3.07.

The firmware and manifest rollback is in #1311.

Tests: `pnpm run build`, `pnpm test`, and `pnpm run check`.


Koenkk:
Thanks!

PR #1305, opened 2026-09-16
Update Inovelli VZM31-SN and VZM32-SN firmware

JamieMagee:
Adds VZM31-SN 3.07 and the VZM32-SN 1.01 downgrade image from Inovelli's firmware repository. After merge, the OTA workflow will make VZM31-SN 3.07 current and move VZM32-SN 1.01 into `images1`.

The autodl change is in #1304.

```json
[
    {
        "fileName": "VZM31-SN_ota_V307_20260326.ota",
        "originalUrl": "https://raw.githubusercontent.com/InovelliUSA/Firmware/main/Blue-Series/Zigbee/VZM31-SN-2-1-Switch/Beta/3.07/VZM31-SN_ota_V307_20260326.ota"
    },
    {
        "fileName": "VZM32-SN_1.01.ota",
        "originalUrl": "https://raw.githubusercontent.com/InovelliUSA/Firmware/main/Blue-Series/Zigbee/VZM32-SN-MMWave-Switch/Beta/1.01/VZM32-SN_1.01.ota"
    }
]
```

PR #1311, opened 2026-09-18
Revert Inovelli beta firmware update

JamieMagee:
Reverts #1305 and its generated manifest changes. This restores VZM31-SN 3.04 as the current release, 2.18 as its previous image, and VZM32-SN 0.10 as the previous image.

The autodl rollback is in #1310.

This PR needs the `ignore-ota-workflow` label because it restores image files and their generated manifest entries together.

Tests: `pnpm test`. The Inovelli manifests and SHA-512 hashes were also checked against the restored files.


JamieMagee:
@Koenkk Could you add the `ignore-ota-workflow` label? This revert must restore the image files and their generated manifest entries in the same PR, so the normal image-only check will reject it.

Koenkk:
Thanks!

